PR #6, opened 2020-06-12
Added grunt banner plugin for copyright banner and restructured License and Copyright related files as per the Fluid Community's conventions

dmahajan980:
I have added one template for copyright boilerplate in the `templates` directory. The current grunt banner configuration that I have set up doesn't require multiple templates (for JavaScript, JSON5, and Markdown files) as long as the content of the boilerplate is the same for different categories of files.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -1,20 +1,60 @@
+/*
+Copyright (c) 2020 The Gamepad Navigator Authors
+See the AUTHORS.md file at the top-level directory of this distribution and at
+https://github.com/fluid-lab/gamepad-navigator/raw/master/AUTHORS.md.
+
+Licensed under the BSD 3-Clause License. You may not use this file except in
+compliance with this License.
+
+You may obtain a copy of the BSD 3-Clause License at
+https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
+*/
+
 /* eslint-env node */
 "use strict";
 
 module.exports = function (grunt) {
     grunt.config.init({
+        licenseBanner: grunt.file.read("templates/LICENSE-banner.txt"),
+        currentMarkdownBanner: /^(<!--)\s[\w\W]+?(-->)\s\s/m,
+        currentJsAndJson5Banner: /^\/\*\sCopyright \(c\)[\w\W]+?\s\*\/\s\s/m,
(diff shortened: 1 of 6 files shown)


Gruntfile.js
```diff
@@ -1,20 +1,60 @@
...
+You may obtain a copy of the BSD 3-Clause License at
+https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
+*/
+
 /* eslint-env node */
 "use strict";
 
 module.exports = function (grunt) {
     grunt.config.init({
+        licenseBanner: grunt.file.read("templates/LICENSE-banner.txt"),
```
the-t-in-rtf, line 18: Unless it causes problems with the plugin, I'd rather see these convenience variables moved under the `usebanner` element, so that the definitions and the references are all together.

Gruntfile.js
```diff
@@ -1,20 +1,60 @@
...
+                md:    ["./*.md", "tests/**/*.md", "docs/*.md"],
                 json:  ["./*.json", "./.*.json", "tests/**/*.json"],
                 json5: ["./*.json5", "tests/**/*.json5"],
                 other: ["./.*"]
             }
+        },
+        usebanner: {
+            jsAndJson5: {
+                options: {
+                    position: "top",
```
the-t-in-rtf, line 33: This and the other position variable can either be omitted (as they are the default) or moved to `usebanner.options` rather than repeated under both tasks.

Gruntfile.js
dmahajan980: I tried moving these variables down and the plugin still worked as before.

Gruntfile.js
dmahajan980: Thanks for the suggestion @the-t-in-rtf. I forgot that `top` is the default value for the banner, so I'll remove this.

dmahajan980:
Hi @the-t-in-rtf, I have modified the `Gruntfile.js` configuration as you suggested. Please review it.

the-t-in-rtf:
I think this is ready to merge, and I'll do so shortly.  Longer term we'll have to talk about how we ensure that the banners are updated regularly as files are added, but I guess we'll have better ideas on that once you start adding files.

PR #8, opened 2020-06-16
Added grunt plugins to generate an unpacked extension and created an infusion component to read gamepad data and track gamepad connection events

dmahajan980:
- Added [grunt-contrib-copy](https://www.npmjs.com/package/grunt-contrib-copy) for copying the `src` files and the minified infusion package from `node_modules` into the `dist` directory which will contain the unpacked version of our chrome extension.
- [grunt-contrib-clean](https://www.npmjs.com/package/grunt-contrib-clean) was added to clear any previously existing files in the `dist` folder before the new unpacked extension files are copied.
- The infusion component would detect the gamepad connected and disconnected events. If connected, it will read gamepad data using the gamepad API as per the frequency (*set to 100ms*).

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,2 @@
+dist/**
+node_modules/**
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,6 @@ report.[0-9]*.[0-9]*.[0-9]*.[0-9]*.json
 
 # Optional eslint cache
 .eslintcache
+
+# Unpacked chrome extension files
+dist/
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -44,15 +44,37 @@ module.exports = function (grunt) {
                 }
             }
         },
+        clean: {
+            previousBuild: {
+                src: ["dist/"]
+            }
+        },
+        copy: {
+            infusion: {
+                expand: true,
+                flatten: true,
+                src: "node_modules/infusion/dist/infusion-all.min.js",
+                dest: "dist/js/lib/infusion/"
+            },
+            source: {
+                expand: true,
+                cwd: "src",
+                src: "**",
+                dest: "dist/"
+            }
+        },
         licenseBanner: grunt.file.read("templates/LICENSE-banner.txt"),
         currentMarkdownBanner: /^(<!--)\s[\w\W]+?(-->)\s\s/m,
         currentJsAndJson5Banner: /^\/\*\sCopyright \(c\)[\w\W]+?\s\*\/\s\s/m
     });
     grunt.loadNpmTasks("gpii-grunt-lint-all");
     grunt.loadNpmTasks("grunt-banner");
+    grunt.loadNpmTasks("grunt-contrib-clean");
+    grunt.loadNpmTasks("grunt-contrib-copy");
 
     grunt.registerTask("lint", "Perform all standard lint checks.", ["lint-all"]);
     grunt.registerTask("banner", "Add copyright banner at the top of files.", ["usebanner"]);
+    grunt.registerTask("build", "Build an unpacked extension.", ["clean", "copy"]);
 
-    grunt.registerTask("default", ["lint"]);
+    grunt.registerTask("default", ["build"]);
 };
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -25,10 +25,15 @@ module.exports = function (grunt) {
             }
         },
         usebanner: {
+            options: {
+                licenseBanner: grunt.file.read("templates/LICENSE-banner.txt"),
+                currentMarkdownBanner: /^(<!--)\s[\w\W]+?(-->)\s\s/m,
+                currentJsAndJson5Banner: /^\/\*\sCopyright \(c\)[\w\W]+?\s\*\/\s\s/m
+            },
             jsAndJson5: {
                 options: {
-                    replace: "<%= currentJsAndJson5Banner %>",
-                    banner: "/*\n<%= licenseBanner %>*/\n"
+                    replace: "<%= usebanner.options.currentJsAndJson5Banner %>",
+                    banner: "/*\n<%= usebanner.options.licenseBanner %>*/\n"
                 },
                 files: {
                     src: ["./src/js/**/*.js", "tests/js/**/*.js", "./*.js", "./*.json5", "tests/**/*.json5"]
@@ -36,8 +41,8 @@ module.exports = function (grunt) {
             },
             markdown: {
                 options: {
-                    replace: "<%= currentMarkdownBanner %>",
-                    banner: "<!--\n<%= licenseBanner %>-->\n"
+                    replace: "<%= usebanner.options.currentMarkdownBanner %>",
+                    banner: "<!--\n<%= usebanner.options.licenseBanner %>-->\n"
                 },
                 files: {
                     src: ["./*.md", "tests/**/*.md", "docs/*.md"]
@@ -62,10 +67,7 @@ module.exports = function (grunt) {
                 src: "**",
                 dest: "dist/"
             }
-        },
-        licenseBanner: grunt.file.read("templates/LICENSE-banner.txt"),
-        currentMarkdownBanner: /^(<!--)\s[\w\W]+?(-->)\s\s/m,
-        currentJsAndJson5Banner: /^\/\*\sCopyright \(c\)[\w\W]+?\s\*\/\s\s/m
+        }
     });
     grunt.loadNpmTasks("gpii-grunt-lint-all");
     grunt.loadNpmTasks("grunt-banner");
@@ -76,5 +78,5 @@ module.exports = function (grunt) {
     grunt.registerTask("banner", "Add copyright banner at the top of files.", ["usebanner"]);
     grunt.registerTask("build", "Build an unpacked extension.", ["clean", "copy"]);
 
-    grunt.registerTask("default", ["build"]);
+    grunt.registerTask("default", ["lint", "banner", "build"]);
 };
--- a/src/js/content_scripts/gamepad-navigator.js
+++ b/src/js/content_scripts/gamepad-navigator.js
@@ -38,7 +38,7 @@ https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
         invokers: {
             attachListener: {
                 funcName: "gamepad.navigator.attachListener",
-                args: ["{that}"]
+                args: ["{that}", "{window}"]
             },
             onConnected: {
                 funcName: "gamepad.navigator.onConnected",
@@ -46,7 +46,7 @@ https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
             },
             onDisconnected: {
                 funcName: "gamepad.navigator.onDisconnected",
-                args: ["{that}", "{arguments}.0"]
+                args: ["{that}"]
             }
         }
     });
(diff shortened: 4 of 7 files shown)


dmahajan980:
@the-t-in-rtf it would be really great if you can review it and suggest changes in the component's structure before we proceed with the navigation part.

Thanks :)

.eslintignore
```diff
@@ -0,0 +1,2 @@
+dist/**
```
the-t-in-rtf, line 1: Why are you not linting "dist"?  What kinds of failures result if you lint that?

.gitignore
```diff
@@ -15,3 +15,6 @@ report.[0-9]*.[0-9]*.[0-9]*.[0-9]*.json
 
 # Optional eslint cache
 .eslintcache
+
+# Unpacked chrome extension files
```
the-t-in-rtf, line 19: Never mind, I see now.

Gruntfile.js
```diff
@@ -44,15 +44,37 @@ module.exports = function (grunt) {
...
+                dest: "dist/js/lib/infusion/"
+            },
+            source: {
+                expand: true,
+                cwd: "src",
+                src: "**",
+                dest: "dist/"
+            }
+        },
         licenseBanner: grunt.file.read("templates/LICENSE-banner.txt"),
```
the-t-in-rtf, line 66: I thought you had moved these variables in the banner plugin's section in another pull?

src/js/content_scripts/gamepad-navigator.js
```diff
@@ -0,0 +1,119 @@
...
+            if (!fluid.isValue(that.model.index)) {
+                // Scan the state of gamepad frequently.
+                that.options.connectivityIntervalReference = setInterval(function () {
+                    // Catch the current state of gamepad.
+                    var connectedGamepad = navigator.getGamepads()[0];
+
+                    // Initiate the gamepad navigator model transaction.
+                    var modelUpdateTransaction = that.applier.initiate();
+
+                    modelUpdateTransaction.fireChangeRequest({ path: "index", value: connectedGamepad.index });
```
the-t-in-rtf, line 65: What happens if you have more than one controller?  I guess your code ignores the second one?

src/js/content_scripts/gamepad-navigator.js
```diff
@@ -0,0 +1,119 @@
...
+             * as an array of floating numbers. To obtain the primary values, we need to
+             * access both differently.
+             */
+            outputObject[index] = fluid.isPrimitive(inputData) ? inputData : inputData.value;
+        });
+        return outputObject;
+    };
+
+    // Create an instance of the gamepad navigator.
+    gamepad.gamepadNavigator();
```
the-t-in-rtf, line 118: I'm happy for this to just be `gamepad.navigator`.

src/js/content_scripts/gamepad-navigator.js
```diff
@@ -0,0 +1,119 @@
...
+
+    /**
+     *
+     * Listens to the connected and disconnected events of the gamepad and updates the
+     * gamepad navigator component's model according to the state of the gamepad.
+     *
+     * @param {Object} that - The gamepad navigator component.
+     *
+     */
+    gamepad.gamepadNavigator.gamepadListener = function (that) {
```
the-t-in-rtf, line 52: I think this should be split out into two functions, one to attach the listener to the window, the other to actually handle the event.  You could also argue for making a wrapper for all gamepad events and writing listeners for each type of event.  So you'd have an "attachListeners" function that ran on creation, and that adds a (window) event listener to fire the associated component event.  Then your "gamepadconnected" listener would be a normal Fluid component listener definition, which could be modified or replaced using configuration options.

src/js/content_scripts/gamepad-navigator.js
```diff
@@ -0,0 +1,119 @@
...
+                    // Commit the current state of gamepad.
+                    modelUpdateTransaction.commit();
+                }, that.options.frequency);
+            }
+        });
+
+        // When gamepad is disconnected
+        window.addEventListener("gamepaddisconnected", function (event) {
+            // Ensure that the gamepad navigator doesn't break if any other gamepad is disconnected.
+            if (that.model.index === event.gamepad.index) {
```
the-t-in-rtf, line 78: So, if someone wants to connect a different controller, they have to ensure that the first controller is detected as disconnected first, and then connect the second controller.  Otherwise they've already missed their chance to use the second controller.  It seems to me like you're probably going to want to handle this differently, either handling multiple controllers separately, or handling inputs from all detected controllers.  One thought I had was to create a dynamic component for each controller when it's connected, and then to destroy that when the controller is detected as having disconnected.

.eslintignore
dmahajan980: Basically the files in this directory are copied from the `src` directory itself which we already lint. Moreover, this directory is not committed to the repository so we don't necessarily need to do it. However, if you think that we should still lint it for some reason, please let me know.

Gruntfile.js
dmahajan980: As I could recall, I thought we had to move these variables below the task definitions. So I added these new task definitions below the banner task and above the convenience variables.

Perhaps I don't understand if you mean to move the variables somewhere else. Could you please clarify on the same?

src/js/content_scripts/gamepad-navigator.js
dmahajan980: I kept it on purpose as I thought that we don't want any second controller to interfere with the primary controller.

dmahajan980: Sorry, just saw your other comment on this approach. I'll reply to it shortly, after reading it.

src/js/content_scripts/gamepad-navigator.js
dmahajan980: Yes, I think that's better and also convenient for reference to this component. I'll make this change in my next commit.

src/js/content_scripts/gamepad-navigator.js
dmahajan980: The second approach you suggested sounds quite interesting to me and I think providing the option to reconfigure would be good for other developers to extend the work as well as for maintaining the project in the long term. I am ready to proceed with the same if you allow me.

src/js/content_scripts/gamepad-navigator.js
dmahajan980: Hi Tony, I started thinking of the flaw in my approach after I read this comment. As you said, in this approach, the user will miss the chance to use the second controller unless they connect it after disconnecting the first controller. So I tried a different approach.

When the user disconnects the primary controller they're using, the extension will get the list of gamepads and the first connected gamepads it finds in the list will be used to serve as the new primary controller. I think with this approach, we won't need to handle multiple controllers and their inputs at once. At the same time, we are offering the users the flexibility to immediately and automatically switch to another gamepad if their current controller fails.

I was able to do this with a bit of restructuring in the current code. Moreover, I think it'll also work as we restructure the gamepad event listener using the approach you suggested in the previous comment. Please let me know what you think and I'll carry this out as we will finalize.

dmahajan980: So I tried to set up the gamepad events wrapper along with the approach I mentioned above, and it seems to work as expected. I can update the pull with that if my approach (mentioned in the previous comment) sounds good to you.

src/js/content_scripts/gamepad-navigator.js
the-t-in-rtf: Thanks.  I think that supporting multiple controllers is really helpful, and it shouldn't be too hard IMO.

.eslintignore
the-t-in-rtf: No, I figured out what you were doing and why later on, it's fine.  As long as the code you write lints, I'm happy.

the-t-in-rtf:
As we don't have tests yet, I'd like to look at your work in progress in a browser.  Could you add basic instructions on installing the extension to the README?

dmahajan980:
Good idea @the-t-in-rtf. I'm on it.

the-t-in-rtf:
It was trivial to figure out how to do it and I did it myself, but your README should generally answer a few key questions:

1. What is the code meant to do?
2. How can I use it?

Eventually you'll add stuff like "How do I run the tests?" as well.

dmahajan980:
@the-t-in-rtf I have a doubt regarding the second question. Apart from mentioning the installation process, won't it be better if we answer its usage after we start to add some navigation features or actions? Like telling them how can they use the default configuration to navigate when we have at least one navigation feature or action that they can use.

the-t-in-rtf:
Yes, in this case, by "use" I did mean "install", and it's fine to keep it to that for now.

dmahajan980:
Thanks for clarifying @the-t-in-rtf. I have added a few lines describing what the gamepad-navigator is and what it can do. I also added instructions along with it. Kindly review it when you have time :)

Gruntfile.js
dmahajan980: @the-t-in-rtf, I tried moving these variables under the `useBanner` block, but sadly, it shows some warnings when used. I think it interferes with the `useBanner`'s configuration. So the best thing we can do is to keep these definitions alongside (either above or below) the `useBanner` task.

src/manifest.json
```diff
@@ -0,0 +1,11 @@
+{
+    "name": "Gamepad Navigator",
+    "description": "The Gamepad Navigator is a configurable application that allows you to navigate web pages and browsers using a game controller.",
```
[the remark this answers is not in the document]
dmahajan980: Sorry, I should have been more careful here. Thanks for noticing this.

Gruntfile.js
the-t-in-rtf: I'm fine with that.

src/js/content_scripts/gamepad-navigator.js
```diff
@@ -0,0 +1,197 @@
...
+     * @param {Object} that - The gamepad navigator component.
+     * @param {Object} event - The window's "gamepaddisconnected" event.
+     *
+     */
+    gamepad.navigator.onDisconnected = function (that, event) {
+        /**
+         * Ensure that the gamepad navigator doesn't break if any other gamepad is
+         * disconnected.
+         */
+        if (that.model.index === event.gamepad.index) {
```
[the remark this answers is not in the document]
the-t-in-rtf: Nothing to be sorry about, one of the points of pull requests is to catch things that a single person would miss on their own.  I have missed things like this many many times in my own refactors.

dmahajan980:
@the-t-in-rtf I've made the changes you suggested. Please check it.

Thanks :)

Gruntfile.js
```diff
@@ -25,19 +25,24 @@ module.exports = function (grunt) {
...
+                licenseBanner: grunt.file.read("templates/LICENSE-banner.txt"),
+                currentMarkdownBanner: /^(<!--)\s[\w\W]+?(-->)\s\s/m,
+                currentJsAndJson5Banner: /^\/\*\sCopyright \(c\)[\w\W]+?\s\*\/\s\s/m
+            },
             jsAndJson5: {
                 options: {
-                    replace: "<%= currentJsAndJson5Banner %>",
-                    banner: "/*\n<%= licenseBanner %>*/\n"
+                    replace: "<%= usebanner.options.currentJsAndJson5Banner %>",
+                    banner: "/*\n<%= usebanner.options.licenseBanner %>*/\n"
```
the-t-in-rtf, line 36: Great!  Glad that worked out.

PR #9, opened 2020-06-22
Added horizontal and vertical scrolling for webpages

dmahajan980:
Along with the scroll feature, this pull introduces a new component `gamepad.inputMapper` which uses the `gamepad.navigator` grade and maps the input values to the navigation functions/actions. A separate namespace called `gamepad.inputMapperUtils` will be the storehouse of these navigation functions/actions and is accessed by `gamepad.inputMapper` to link those functions with inputs.


dmahajan980:
@the-t-in-rtf kindly review it when you have time. After it gets merged, we can hopefully go for the tabbing part of the project.

Thanks

src/js/content_scripts/input-mapper-utils.js
```diff
@@ -0,0 +1,182 @@
...
+    "use strict";
+
+    fluid.registerNamespace("gamepad.inputMapperUtils");
+
+    /**
+     *
+     * Scroll horizontally across the webpage.
+     *
+     * @param {Object} that - The inputMapper component.
+     * @param {Integer} value - The value of the gamepad input.
```
the-t-in-rtf, line 26: It would be good to describe the units here.  Is it pixels?

You also need to give some thought to how this will work with analogue controls like thumbsticks, as they will keep sending values very rapidly.  So the net effect is that scrolling will be very fast on a thumbstick or other analogue control.  On the flip side, you need some way to ensure that a digital button can be used reasonably without requiring a hundred button presses.

One idea would be to handle axes differently, since those send repeated values by their nature.  You might also make the scale adjustable per button, which would help with triggers as well.

src/js/content_scripts/input-mapper-utils.js
```diff
@@ -0,0 +1,182 @@
...
+         * Stop scrolling for the previous input value. Also stop scrolling if the input
+         * source (analog/button) is at rest.
+         */
+        clearInterval(that.options.intervalRecord.leftScroll);
+
+        /**
+         * Scroll the webpage towards the left only if the input value is more than the
+         * specified value.
+         */
+        if (value > 0.20) {
```
the-t-in-rtf, line 65: This cutoff should be an option.

src/js/content_scripts/input-mapper-utils.js
```diff
@@ -0,0 +1,182 @@
...
+
+    /**
+     *
+     * Scroll the webpage in downward direction.
+     *
+     * @param {Object} that - The inputMapper component.
+     * @param {Integer} value - The value of the gamepad input.
+     *
+     */
+    gamepad.inputMapperUtils.scrollDown = function (that, value) {
```
the-t-in-rtf, line 163: I think you should have a single function with an "axis" argument rather than having these two very similar functions.

src/js/content_scripts/input-mapper-utils.js
```diff
@@ -0,0 +1,182 @@
...
+        /**
+         * Stop scrolling for the previous input value. Also stop scrolling if the input
+         * source (analog/button) is at rest.
+         */
+        clearInterval(that.options.intervalRecord.downwardScroll);
```
the-t-in-rtf, line 168: I would suggest adding an `onDestroy` listener to ensure that the intervals are cleared if the component is destroyed.

I also am surprised that you were able to get away with storing information that presumably changes in options.  Later versions of Infusion will reject this outright.  It's better to use `members` for non-model variable data like this.

src/js/content_scripts/input-mapper-utils.js
```diff
@@ -0,0 +1,182 @@
...
+
+        /**
+         * Scroll the webpage downward only if the input value is more than the specified
+         * value.
+         */
+        if (value > 0.20) {
+            // Scroll upward according to the new input value.
+            that.options.intervalRecord.downwardScroll = setInterval(function () {
+                var yOffset = $(window).scrollTop();
```
the-t-in-rtf, line 177: I would suggest giving your local `window` variable another name like `windowObject` to ensure that there are no side effects of clobbering the global `window` variable.

src/js/content_scripts/input-mapper.js
```diff
@@ -0,0 +1,128 @@
...
+            downwardScroll: null,
+            leftScroll: null,
+            rightScroll: null
+        },
+        invokers: {
+            produceNavigation: {
+                funcName: "gamepad.inputMapper.produceNavigation",
+                args: ["{that}", "{arguments}.0"]
+            },
+            scrollHorizontally: {
```
the-t-in-rtf, line 73: I see now, you have separate methods for button and analogue inputs.  That makes sense.

dmahajan980: Yes, basically, we have `scrollLeft`, `scrollRight`, `scrollUp`, and `scrollDown` for buttons and triggers, while at the same time, `scrollHorizontally` and `scrollVertically` for axes (which use the former functions).

dmahajan980: I've written these in such a manner, keeping in mind that the user could reconfigure the controls at a later point in time.

src/js/content_scripts/input-mapper.js
```diff
@@ -0,0 +1,128 @@
...
+
+    var gamepad = fluid.registerNamespace("gamepad");
+    fluid.registerNamespace("gamepad.inputMapperUtils");
+
+    fluid.defaults("gamepad.inputMapper", {
+        gradeNames: ["gamepad.navigator"],
+        model: {
+            actionMap: {
+                buttons: {
+                    "0": "vacant",
```
the-t-in-rtf, line 26: I guess you're thinking that you'd have separate maps for the action and any additional options, but to me it makes more sense to have an object that can be used to control the action and also specify non-default values relevant to the action.  For example, you might want to indicate that a thumbstick is to be used to scroll, which axis, how quickly, and whether to invert the direction of the scrolling.  It seems easier to do this with a single object per control rather than separate maps.

src/js/content_scripts/input-mapper.js
```diff
@@ -0,0 +1,128 @@
...
+         */
+        var inputType = change.path[0],
+            index = change.path[1],
+            inputValue = change.value;
+
+        // Retrieve the label for the action to be triggered on gamepad disturbance.
+        var actionLabel = that.model.actionMap[inputType][index];
+
+        // Execute the navigation effect if the button is not set to "vacant".
+        if (actionLabel !== "vacant") {
```
the-t-in-rtf, line 122: I don't really like strings that have special meanings like this.  If you have a property that describes the action, setting that property to `false` gives more the correct impression IMO.

src/js/content_scripts/input-mapper.js
```diff
@@ -0,0 +1,128 @@
...
+            fluid.get(that, actionLabel)(inputValue);
```
the-t-in-rtf, line 123: You really need to confirm that `fluid.get(that, actionLabel)` returns a function.  If (for example) `actionLabel` has a typo, you'll be trying to execute `undefined` with arguments.

src/js/content_scripts/input-mapper.js
```diff
@@ -0,0 +1,128 @@
...
+https://github.com/fluid-lab/gamepad-navigator/raw/master/AUTHORS.md.
+
+Licensed under the BSD 3-Clause License. You may not use this file except in
+compliance with this License.
+
+You may obtain a copy of the BSD 3-Clause License at
+https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
+*/
+
+/* eslint-env browser */
```
the-t-in-rtf, line 13: I would suggest setting this in your `.eslintrc.json` so that you only need to specify the environment for things like the `Gruntfile.js` that run in Node.

src/js/content_scripts/input-mapper-utils.js
the-t-in-rtf: Disregard my comments about the analogue controls, I see how you handled that.  You should still describe the units.

src/js/content_scripts/input-mapper.js
```diff
@@ -0,0 +1,128 @@
...
+                    "13": "scrollDown",
+                    "14": "scrollLeft",
+                    "15": "scrollRight",
+                    "16": "vacant"
+                },
+                axes: {
+                    "0": "scrollHorizontally",
+                    "1": "scrollVertically",
+                    "2": "vacant",
+                    "3": "vacant"
```
the-t-in-rtf, line 48: Although the standard reference controller in the [Gamepad API standard](https://www.w3.org/TR/gamepad/) has 4 axes, there are controllers with more (or less axes).  This is one way you might end up trying to launch `undefined` as a function using your current approach.  I think for now simply confirming that you were able to find an action mapped to the input is enough to prevent this type of problem.

src/js/content_scripts/input-mapper.js
dmahajan980: Hi @the-t-in-rtf, thank you for the feedback. I haven't really thought of separate maps and I named it to `actionMaps` as that's the only option we are dealing with as of now.

I'll change it as you suggested, however, I am slightly confused here. I just need a heads-up if what I understood is correct. Did you mean that each button/axes should be an object and would contain options such as `defaultAction` and `currentAction` (for reconfigured value)? Also, if this is correct, what other options would you want me to include?

dmahajan980: Nevermind, I think I got the idea. But I am not exactly sure if what I've done is exactly what you meant here. I guess you would have to review my next commit to see if that's correct.

dmahajan980:
@the-t-in-rtf kindly review the pull request. I've updated it according to your feedback. Pardon me for putting such a gigantic pull request for you. I will try to keep these pulls as short as I can from now on.

Thank you

src/js/content_scripts/gamepad-navigator.js
```diff
@@ -186,4 +189,17 @@ https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
...
+    /**
+     *
+     * A listener for the gamepad navigator component to clear the connectivity interval
+     * when the instance of the component is destroyed.
+     *
+     * @param {Number} connectivityIntervalReference - The reference to the gamepad
+     *                                                 connection interval loop.
+     *
+     */
+    gamepad.navigator.clearConnectivityInterval = function (connectivityIntervalReference) {
```
the-t-in-rtf, line 202: Unless you're doing something like checking to make sure the interval is non-empty, you should be able just to use `funcName: "clearInterval"` in an invoker or listener definition and should not need to write a function just for this.

the-t-in-rtf:
Just tried it out with a controller, this is a very good start.  If you can tidy up that unnecessary function in favour of an invoker, I'm happy to merge.

dmahajan980:
Thank you @the-t-in-rtf. I'll put the changes in a minute for you :)

dmahajan980:
@the-t-in-rtf kindly check again.

PR #10, opened 2020-06-25
Added tabbing and reverse tabbing feature

dmahajan980:


src/js/content_scripts/input-mapper.js
```diff
@@ -42,14 +42,14 @@ https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
                         speedFactor: 1
                     },
                     "4": {
```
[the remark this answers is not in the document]
the-t-in-rtf: i.e. "Left Bumper"

the-t-in-rtf:
As we discussed in chat, this works with GitHub, but doesn't work with sites like [The gamepad API tester](https://html5gamepad.com/).

I also have encountered multiple pages where I sometimes have to issue multiple clicks to navigate, such as 

* [https://developer.mozilla.org/en-US/docs/Web/API/Gamepad_API/Using_the_Gamepad_API](https://developer.mozilla.org/en-US/docs/Web/API/Gamepad_API/Using_the_Gamepad_API)
* [Google search results](https://www.google.com/search?q=gamepad+api&rlz=1C5CHFA_enNL878NL878&oq=gamepad+api&aqs=chrome..69i57j0l6j69i60.3152j0j7&sourceid=chrome&ie=UTF-8)

src/js/content_scripts/input-mapper.js
```diff
@@ -124,9 +124,9 @@ https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
                         invert: false
                     },
                     "3": {
-                        defaultAction: null,
+                        defaultAction: "analogTabbing",
```
the-t-in-rtf, line 127: The analog tabbing should probably be bound to buttons 6 and 7, which are the only analog controls on PS4 dualshock and Xbox One X controllers.

the-t-in-rtf: I'd also ask that you put the binding definitions in numeric order, it makes reviewing the list easier.

the-t-in-rtf:
I can only verify buttons 4 and 5 working.  I cannot verify that the analog controls work with the digital buttons on my gamepad (see comment about remapping the analog tabbing to use buttons 6 and 7 for now).

src/js/content_scripts/input-mapper-utils.js
```diff
@@ -188,4 +188,156 @@ https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
...
     };
+
+    /**
+     *
+     * Tab through the webpage using analog sticks.
+     *
+     * @param {Object} that - The inputMapper component.
+     * @param {Integer} value - The value of the gamepad input.
+     * @param {Integer} speedFactor - Times by which the tabbing speed should be increased.
+     * @param {Boolean} invert - Determine if the tabbing should be in opposite order.
```
the-t-in-rtf, line 199: I think "Whether" would be better than "Determine if" here.

src/js/content_scripts/input-mapper-utils.js
```diff
@@ -188,4 +188,156 @@ https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
...
+     *
+     */
+    gamepad.inputMapperUtils.analogTabbing = function (that, value, speedFactor, invert) {
+        var inversionFactor = invert ? -1 : 1;
+        value = value * inversionFactor;
+        if (value > 0.20) {
```
the-t-in-rtf, line 205: As with other cuttoff values, the bounds should be configurable.

src/js/content_scripts/input-mapper-utils.js
```diff
@@ -188,4 +188,156 @@ https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
...
+     * @param {Object} that - The inputMapper component.
+     * @param {Integer} value - The value of the gamepad input.
+     * @param {Integer} speedFactor - Times by which the tabbing speed should be increased.
+     *
+     */
+    gamepad.inputMapperUtils.forwardTab = function (that, value, speedFactor) {
+        // Stop tabbing for the previous input value.
+        clearInterval(that.options.members.intervalRecords.forwardTab);
+
+        if (value > that.options.cutoffValue) {
```
the-t-in-rtf, line 232: I'm not sure why you have the bounds checking twice for analog controls.  Seems like you should only have it once, and in the analogue function.

src/js/content_scripts/input-mapper-utils.js
```diff
@@ -188,4 +188,156 @@ https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
...
+                }
+                else {
+                    var activeElementIndex = tabbableElements.indexOf(document.activeElement);
+
+                    /**
+                     * If the currently focused element is not found in the list, refer to
+                     * the stored value of the index.
+                     */
+                    if (activeElementIndex === -1) {
+                        activeElementIndex = that.options.members.currentTabIndex;
```
the-t-in-rtf, line 254: This should be `that.currentTabIndex`, the option is only to define the member variable.

src/js/content_scripts/input-mapper.js
```diff
@@ -150,8 +150,11 @@ https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
                 upwardScroll: null,
                 downwardScroll: null,
                 leftScroll: null,
-                rightScroll: null
-            }
+                rightScroll: null,
+                forwardTab: null,
+                reverseTab: null
+            },
+            currentTabIndex: 0
```
the-t-in-rtf, line 157: We may have to leave it for now, but I'm concerned that your approach will have problems with dynamic interfaces, i.e. if the index is stored and the number of tabbable elements changes.  We can add tests to confirm this down the road.

src/js/content_scripts/input-mapper.js
dmahajan980: `analogTabbing` was meant for joysticks, as when the user wants to tab using joystick. For the buttons, I've added `forwardTabbing` and `reverseTabbing`. I think I made a mistake with the function naming. I'll change it then.

src/js/content_scripts/input-mapper-utils.js
dmahajan980: I'm sorry. That was a silly mistake on my part. It should have been 0 instead as the case with joysticks on the gamepad. I'll rectify this.

src/js/content_scripts/input-mapper.js
dmahajan980: This index will be used only under limited conditions. I guess you're thinking that we are using the index every time we tab, but that's not the case. It is kept as a [fallback option](https://github.com/fluid-lab/gamepad-navigator/pull/10/files#diff-cc20e237566e63150ac1d82f0dbee3f8R249) and will be used only if `indexOf` fails under certain circumstances. However, we can confirm this with tests later as we proceed.

dmahajan980:
The latest commit should fix issue #11 and hopefully other issues that are related to it. I've also added the reasoning for the occurrence of the issue in its comment section.

src/js/content_scripts/input-mapper.js
the-t-in-rtf: You're going to have to thing about the problem in general of making actions for:

1. Digital button.
2. Analogue buttons such as a triggers.
3. Axes.

The only difference between analogue buttons and axes is that axes have their own direction.  I think you can (and should) make it possible for analogue buttons to be used for anything that axes can be used for.  After all, if you allow people to remap controls but they can only map certain actions to the axes, that's more limiting than allowing them to assign some actions to analogue buttons.  We can leave a TODO for now, but this will really be important once you have a configuration interface.

the-t-in-rtf:
In general the latest revision works much better.  I can get the cursor to "stick" (not move when I press a button) and "skip" (move two steps when I only press once).  I will look at the code in some depth, but this may be the type of thing we can investigate better when you're writing tests.

src/js/content_scripts/input-mapper.js
dmahajan980: Actually I've been writing the functions for buttons in such a way that we can use it for triggers as well. For example, the `cutoffValue` that I set up at certain places was put in regard to these triggers.

dmahajan980: Moreover, in another light, you can also use the functions written for axes in triggers. If we take this approach, we just have to put an additional `invert` option to their configuration (as we did in axes) so that the user can choose to have the inverted action.

dmahajan980:
@the-t-in-rtf I've added TODOs as a reference to the issue you're facing. I'll investigate it further as we start writing tests. If there's anything else you want me to change before we merge this pull then please let me know.

PR #12, opened 2020-06-29
Added tests for scroll navigation.

dmahajan980:


src/js/content_scripts/gamepad-navigator.js
```diff
@@ -30,7 +30,7 @@ https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
         listeners: {
             onCreate: "{that}.attachListener",
             onGamepadConnected: "{that}.onConnected",
-            onGamepadDisconnected: "{that}.onDisconnected",
+            "onGamepadDisconnected.handleConectedGamepads": "{that}.onDisconnected",
```
the-t-in-rtf, line 33: Should be `handleConnectedGamepads` with two Ns.

src/js/content_scripts/input-mapper.js
```diff
@@ -186,6 +191,7 @@ https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
             currentTabIndex: 0
         },
         cutoffValue: 0.20,
+        multiplicationFactor: 50,
```
the-t-in-rtf, line 194: "multiplication" is a bit generic, it'd be nice to find a better word or phrase to describe and distinguish this from `speedFactor`.

src/js/content_scripts/instance.js
```diff
@@ -0,0 +1,18 @@
+/*
```
the-t-in-rtf, line 1: "instance" is far far too generic.  What is it an "instance" of?  It seems like another input mapper?

tests/js/gamepad-mock.js
```diff
@@ -0,0 +1,105 @@
...
+    /**
+     *
+     * Generate a mock of the standard Gamepad object. Optionally, it can take the
+     * inputSpec as its parameter. The inputSpec should be in the following formats:
+     *
+     * {
+     *   buttons: { "2": 1 }
+     * }
+     *
+     * The above inputSpec will return a gamepad mock with the value of button 2 to 1.
```
the-t-in-rtf, line 29: "set to 1".

tests/js/gamepad-mock.js
```diff
@@ -0,0 +1,105 @@
...
+     *
+     * {
+     *   axes: { "1": 0.753 }
+     * }
+     *
+     * The above inputSpec will return a gamepad mock with the value of axes 2 to 0.753.
```
the-t-in-rtf, line 35: I'd suggest changing "to" to "set to" in all of these.

tests/js/gamepad-mock.js
```diff
@@ -0,0 +1,105 @@
...
+     *
+     * The above inputSpec will return a gamepad mock with the value of button 5 to 1,
+     * button 11 to 1, and of axes 3 to 0.92.
+     *
+     * @param {Object} inputSpec - (optional) The object containing the desired value for
+     *                             the gamepad inputs.
+     * @return {Array} - The array containing the mock gamepad mock object.
+     *
+     */
+    gamepad.tests.utils.generateGamepadMock = function (inputSpec) {
```
the-t-in-rtf, line 53: I'm a little worried that this approach will not be sufficient, it seems to me like you will often need to trigger a change of button state rather than initialising the mock with a particular state.

tests/js/scroll/scroll-bidirectional-tests.js
```diff
@@ -0,0 +1,249 @@
...
+                    var isScrolledLeft = $(window).scrollLeft() < xPosition ? true : false;
+                    jqUnit.assertTrue("The page should have scrolled left.", isScrolledLeft);
+
+                    testNavigator.events.onGamepadDisconnected.fire();
+                    clearInterval(gamepad.tests.intervalLoop);
+                    testNavigator.destroy();
+                    jqUnit.assertTrue("The instance of the gamepad navigator should be destroyed.", fluid.isDestroyed(testNavigator));
+                    jqUnit.start();
+                }
+            }, gamepad.tests.frequency);
```
the-t-in-rtf, line 64: Pitting one interval against another seems like it could easily lead to race conditions once you start trying to test more complex sequences of keys, but as long as the test frequency is significantly longer than the gamepad polling frequency I guess it's fine for now.

tests/js/scroll/scroll-bidirectional-tests.js
```diff
@@ -0,0 +1,249 @@
...
+            jqUnit.assertTrue("The scroll position should be at the top-left of the webpage.", isTopLeft);
+
+            var testNavigator = gamepad.inputMapper({
+                windowObject: gamepad.tests.windowObject,
+                frequency: gamepad.tests.frequency
+            });
+            jqUnit.assertValue("The Gamepad Navigator should be instantiated.", testNavigator);
+
+            // Trigger gamepadconnected event and start scrolling towards bottom-right corner.
+            testNavigator.events.onGamepadConnected.fire();
```
the-t-in-rtf, line 128: I see the pattern you're using, which is to have a gamepad in a permanent state, then connect it, then eventually disconnect it and look at the change.  I'm thinking that we need a test version of whatever component is polling the gamepads that can be manually instructed to poll the gamepads.   We'd also need a gamepad whose state can be updated.  That way you can much more cleanly establish the sequence of events in these tests, i.e.:

1. Scroll the page to the top.
2. Confirm the initial state.
3. Connect a gamepad with no buttons.
4. Instruct the test component to poll the gamepads once.
5. Confirm that the scrolling hasn't changed.
6. Update the gamepad to hold down a control.
7. Instruct the test component to poll the gamepads once.
8. Confirm the scrolling has changed.

You would only need to test the recurring polling mechanism (setInterval) in your end-to-end tests.

package.json
```diff
@@ -13,7 +13,8 @@
     },
     "homepage": "https://github.com/fluid-lab/gamepad-navigator#readme",
     "scripts": {
-        "lint": "grunt lint"
+        "lint": "grunt lint",
+        "test": "testem ci --file tests/testem.json"
```
the-t-in-rtf, line 17: You need to add testem as a dev dependency so that your tests can be run with a particular version of testem both locally and in CI.

tests/testem.json
```diff
@@ -0,0 +1,9 @@
+{
+    "test_page": "tests/all-tests.html",
+    "launch": "Chrome",
+    "browser_args": {
+        "Chrome": [
+            "--disable-extensions-except=dist/"
```
the-t-in-rtf, line 6: I'm a little confused by this line.  From our discussion, sounds like you should just disable all extensions.

the-t-in-rtf:
Just a quick summary of the changes we discussed now that I understand your approach better.

1. We should not be cloning or replacing the global window object.
2. The [polling function](https://github.com/dmahajan980/gamepad-navigator/blob/50d1297e3505148be5d2a4371444ee686a822edc/src/js/content_scripts/gamepad-navigator.js#L94) should be broken out into an invoker that uses `that.options.windowObject.navigator` instead of the global `navigator`.
3. You should create a test instance of the navigator that does not poll the gamepads at all or manage connect and disconnect.   Instead, you will manually call the polling function each time you change the state.
4. Your window object should have its own `navigator` element whose `getGamepads` element points to another test invoker that returns a list of connected gamepads and their state.  The gamepads could in fact simply be represented as member data that you manipulate in your tests before checking the results.

Making the window/navigator mock operate purely synchronously like this should greatly simplify writing tests and also help avoid a whole range of timing issues.

dmahajan980:
@the-t-in-rtf I've made the changes as we discussed. There would be a bit of asynchronicity in our tests as the navigation operations are asynchronous in a sense, but I've tried to simplify those for a much cleaner sequence of operations. Kindly review it and let me know if I still missed something that should've been covered.

Thank you :)

tests/js/gamepad-mock.js
dmahajan980: This function is just an all-purpose mock generator for use in all types of tests in the future. We are using other functions in our tests, which use this function for generating a gamepad mock with varying states (such as change in button values, axes values, etc.).

the-t-in-rtf: We can certainly give it a try and change the approach if it becomes burdensome.

tests/js/scroll/scroll-bidirectional-tests.js
the-t-in-rtf: I think I'm fine with the slightly different approach you used.  There is a small risk that the tests will fail if you connect or disconnect a gamepad while they're running, but I'd rather revisit if we encounter problems doing things this way.  I do think that eventually you'll want a more convenient pattern than writing out the replacement functions, but at least what you have now it does make it fairly clear what you're doing and avoids potential race conditions with competing intervals and timeouts.

PR #13, opened 2020-07-01
Added tests for tab navigation.

dmahajan980:
@the-t-in-rtf Kindly review this pull request. Once we have this ready, we should merge this after merging the scroll-tests pull #12.

### DIFF
--- a/tests/js/tab/tab-button-continuous-tests.js
+++ b/tests/js/tab/tab-button-continuous-tests.js
@@ -19,7 +19,7 @@ https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
 
         fluid.registerNamespace("gamepad.tests");
 
-        jqUnit.module("Gamepad Navigator Button Tab Navigation Tests", {
+        jqUnit.module("Gamepad Navigator Button (Continuous) Tab Navigation Tests", {
             setup: function () {
                 gamepad.tests.windowObject = window;
                 gamepad.tests.count = 2;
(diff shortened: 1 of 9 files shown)


the-t-in-rtf:
Can you merge this with master?  As it stands the package.json lacks the test commands you added there.  I know I can run things manually, but it's nicer if I can confirm everything works in context.

tests/js/tab/tab-order-tests.js
```diff
@@ -0,0 +1,66 @@
...
+            var lastTabbableElement = gamepad.tests.sortedTabbableElements[gamepad.tests.sortedTabbableElements.length - 1],
+                isLastElement = lastTabbableElement.getAttribute("id") === "last";
+
+            jqUnit.assertTrue("The first element should be the last div with id=\"last\"", isLastElement);
+        });
+
+        jqUnit.test("Verify presence and absence of elements according to their tabindex value.", function () {
+            jqUnit.expect(2);
+
+            // Check if the tabbable elements list contains elements with tabindex="-1".
```
the-t-in-rtf, line 55: Very good.

tests/js/tab/tab-button-tests.js
```diff
@@ -0,0 +1,226 @@
...
+                var afterPollingFocusedElementTabIndex = document.activeElement.getAttribute("tabindex");
+
+                // Check if the focus has moved in the forward direction.
+                var hasTabbedForward = beforePollingFocusedElementTabIndex < afterPollingFocusedElementTabIndex;
+                jqUnit.assertTrue("The focus should have moved in the forward direction.", hasTabbedForward);
+                jqUnit.start();
+            }, gamepad.tests.frequency * 4);
+        });
+
+        jqUnit.asyncTest("Change the focus to one of the previous elements in reverse tabbing using buttons.", function () {
```
the-t-in-rtf, line 182: I think it would be better to be more specific than "button".  I would suggest using "continuous" and "discrete".  "Continuous" would be for held buttons where the action continues.  "Discrete" would be for activities that happen once when a button is pressed and released.  Once we get to the point where you have a control panel, you'll want to let people choose which style of navigation to use.  For example, people with limited motor control may need "discrete" controls to avoid triggering an action twice before they can release the button.

tests/js/tab/tab-button-continuous-tests.js
```diff
@@ -0,0 +1,142 @@
...
+            // Wait for a few milliseconds for the navigator to change focus.
+            setTimeout(function () {
+                // Restore the gamepad back to its neutral state.
+                gamepad.tests.navigator.pollGamepads();
+
+                // Record the index of the element currently focused.
+                var afterPollingFocusedElementTabIndex = document.activeElement.getAttribute("tabindex");
+
+                // Check if the focus has moved in the forward direction.
+                var hasTabbedForward = beforePollingFocusedElementTabIndex < afterPollingFocusedElementTabIndex;
```
the-t-in-rtf, line 92: Smart and pretty robust against most timing issues.

the-t-in-rtf:
As we discussed, I'm still seeing issues where multiple button presses are required to tab, but I'll open a separate ticket for that.

PR #16, opened 2020-07-02
Added click feature and related tests

dmahajan980:
This approach uses a simple `click` handler and special handling for `select` dropdown menus. Along with this, I have added those tests which don't involve navigation between pages or tab switching. For these kinds of tests, we'll probably use the WebDriver or some other tool in the future, as we discussed in our chat.

### DIFF
--- a/src/js/content_scripts/input-mapper-utils.js
+++ b/src/js/content_scripts/input-mapper-utils.js
@@ -349,4 +349,49 @@ https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
             return 0;
         }
     };
+
+    /**
+     *
+     * Click on the currently focused element.
+     *
+     * @param {Integer} value - The value of the gamepad input.
+     *
+     */
+    gamepad.inputMapperUtils.click = function (value) {
+        if (value > 0) {
+            /**
+             * If SELECT element is currently focused, toggle its state. Otherwise perform
+             * the regular click operation.
+             */
+            if (document.activeElement.nodeName === "SELECT") {
+                var optionsLength = 0;
+
+                // Compute the number of options and store it.
+                document.activeElement.childNodes.forEach(function (childNode) {
+                    if (childNode.nodeName === "OPTION") {
+                        optionsLength++;
+                    }
+                });
+
+                // Toggle the SELECT dropdown.
+                if (!document.activeElement.getAttribute("size") || document.activeElement.getAttribute("size") === "1") {
+                    /**
+                     * Allow limited expansion to avoid an overflowing list, considering the
+                     * list could go as large as 100 or more (for example, a list of
+                     * countries)
+                     */
+                    var length = optionsLength < 15 ? optionsLength : 15;
+                    document.activeElement.setAttribute("size", length);
+                }
+                else {
+                    // Restore the size of the dropdown.
+                    document.activeElement.setAttribute("size", 1);
+                }
+            }
+            else {
+                // Click on the focused element.
+                document.activeElement.click();
+            }
+        }
+    };
 })(fluid, jQuery);
--- a/src/js/content_scripts/input-mapper-utils.js
+++ b/src/js/content_scripts/input-mapper-utils.js
@@ -351,6 +351,14 @@ https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
     };
 
     /**
+     * TODO: Replace count global variable as members in tests apart from the click tests.
+     * TODO: Replace the initial focus/webpage checks and modelAtRest global variable
+     *       setup with helper functions in tests apart from the click tests.
+     * TODO: Add horizontal rule between the QUnit markup and test-fixture block in tests
+     *       apart from the click tests.
+     */
+
+    /**
      *
      * Click on the currently focused element.
      *
@@ -376,16 +384,28 @@ https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
                 // Toggle the SELECT dropdown.
                 if (!document.activeElement.getAttribute("size") || document.activeElement.getAttribute("size") === "1") {
                     /**
+                     * Store the initial size of the dropdown in a separate attribute
+                     * (if specified already).
+                     */
+                    var initialSizeString = document.activeElement.getAttribute("size");
+                    if (initialSizeString) {
+                        document.activeElement.setAttribute("initialSize", parseInt(initialSizeString));
+                    }
+
+                    /**
                      * Allow limited expansion to avoid an overflowing list, considering the
                      * list could go as large as 100 or more (for example, a list of
-                     * countries)
+                     * countries).
                      */
-                    var length = optionsLength < 15 ? optionsLength : 15;
+                    var length = Math.min(15, optionsLength);
                     document.activeElement.setAttribute("size", length);
                 }
                 else {
+                    // Obtain the initial size of the dropdown.
+                    var sizeString = document.activeElement.getAttribute("initialSize") || "1";
+
                     // Restore the size of the dropdown.
-                    document.activeElement.setAttribute("size", 1);
+                    document.activeElement.setAttribute("size", parseInt(sizeString));
                 }
             }
             else {
--- a/tests/html/click-tests.html
+++ b/tests/html/click-tests.html
@@ -1,0 +1,82 @@
+<!DOCTYPE html>
+<html>
+    <head>
+        <meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
+        <title>Gamepad Navigator Tabbing Tests</title>
+
+        <link rel="stylesheet" media="screen" href="../../node_modules/infusion/tests/lib/qunit/css/qunit.css" />
+        <link rel="stylesheet" href="../../node_modules/infusion/tests/lib/qunit/addons/composite/qunit-composite.css">
+
+        <script type="text/javascript" src="../../node_modules/infusion/dist/infusion-all.min.js"></script>
+        <script type="text/javascript" src="../../node_modules/ally.js/ally.min.js"></script>
+        <script type="text/javascript" src="../../node_modules/infusion/tests/lib/qunit/js/qunit.js"></script>
+        <script type="text/javascript" src="../../node_modules/infusion/tests/test-core/jqUnit/js/jqUnit.js"></script>
+
+        <script type="text/javascript" src="../../src/js/content_scripts/gamepad-navigator.js"></script>
+        <script type="text/javascript" src="../../src/js/content_scripts/input-mapper-utils.js"></script>
+        <script type="text/javascript" src="../../src/js/content_scripts/input-mapper.js"></script>
+
+        <script type="text/javascript" src="../js/gamepad-mock.js"></script>
+        <script type="text/javascript" src="../js/click/click-mock-utils.js"></script>
+        <script type="text/javascript" src="../js/click/click-basic-tests.js"></script>
+        <script type="text/javascript" src="../js/click/click-dropdown-tests.js"></script>
+    </head>
+    <body id="body">
+        <h1 id="qunit-header">Gamepad Navigator Click Tests</h1>
+        <h2 id="qunit-banner"></h2>
+        <div id="qunit-testrunner-toolbar"></div>
+        <h2 id="qunit-userAgent"></h2>
+        <ol id="qunit-tests"></ol>
+
+        <div class="test-elements">
+            <br>
+            <button type="button" onclick="gamepad.tests.clicked = true;">Sample Button</button>
+            <br><br>
+            <fieldset id="radios">
+                <input type="radio" id="radio-one" name="group1" value="Radio One"><label for="radio-one">Radio One</label><br>
+                <input type="radio" id="radio-two" name="group1" value="Radio Two"><label for="radio-two">Radio Two</label>
+            </fieldset><br>
+            <fieldset id="checkboxes">
+                <input type="checkbox" id="checkbox-one" name="group1">
+                <label for="checkbox-one"> Checkbox One</label><br>
+                <input type="checkbox" id="checkbox-two" name="group1">
+                <label for="checkbox-two"> Checkbox Two</label><br>
+                <input type="checkbox" id="checkbox-three" name="group1">
+                <label for="checkbox-three"> Checkbox Three</label>
+            </fieldset><br>
+            <br>
+            <label for="dropdown-short">Short Dropdown: </label>
+            <select id="dropdown-short">
+                <option id="short-one">one</option>
+                <option id="short-two">two</option>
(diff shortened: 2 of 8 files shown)


dmahajan980:
@the-t-in-rtf I would request you to review it.

Thanks

src/js/content_scripts/input-mapper-utils.js
```diff
@@ -349,4 +349,49 @@ https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
...
+                });
+
+                // Toggle the SELECT dropdown.
+                if (!document.activeElement.getAttribute("size") || document.activeElement.getAttribute("size") === "1") {
+                    /**
+                     * Allow limited expansion to avoid an overflowing list, considering the
+                     * list could go as large as 100 or more (for example, a list of
+                     * countries)
+                     */
+                    var length = optionsLength < 15 ? optionsLength : 15;
```
the-t-in-rtf, line 383: Math.max(15, optionsLength) is probably shorter and better here.

src/js/content_scripts/input-mapper-utils.js
```diff
@@ -349,4 +349,49 @@ https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
...
+                    document.activeElement.setAttribute("size", length);
+                }
+                else {
+                    // Restore the size of the dropdown.
+                    document.activeElement.setAttribute("size", 1);
```
the-t-in-rtf, line 388: I would read up just to know to see if there are patterns where people use sizes other than 1.  You might have to keep track of what it used to be somehow, for example by stashing it in an attribute on the selectbox you're opening that you can read on close.

tests/html/click-tests.html
```diff
@@ -0,0 +1,82 @@
+<!DOCTYPE html>
+<html>
+    <head>
+        <meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
+        <title>Gamepad Navigator Tabbing Tests</title>
```
the-t-in-rtf, line 5: This title copied from the other test should be updated.  (I do this all the time myself, it's not a big deal.)

tests/html/click-tests.html
```diff
@@ -0,0 +1,82 @@
...
+        <script type="text/javascript" src="../js/click/click-dropdown-tests.js"></script>
+    </head>
+    <body id="body">
+        <h1 id="qunit-header">Gamepad Navigator Click Tests</h1>
+        <h2 id="qunit-banner"></h2>
+        <div id="qunit-testrunner-toolbar"></div>
+        <h2 id="qunit-userAgent"></h2>
+        <ol id="qunit-tests"></ol>
+
+        <div class="test-elements">
```
the-t-in-rtf, line 31: I was curious about this earlier, did you try putting your test fixtures in the hidden div that QUnit uses?  I sometimes have to avoid using that, just wanted to know if that's also the case here.

tests/js/click/click-basic-tests.js
```diff
@@ -0,0 +1,143 @@
...
+            // Confirm that the instance of the gamepad navigator is created.
+            gamepad.tests.navigator = gamepad.inputMapper({
+                windowObject: gamepad.tests.windowObject
+            });
+            jqUnit.assertTrue("The Gamepad Navigator should be instantiated.", fluid.isComponent(gamepad.tests.navigator));
+
+            // Check the state of gamepad inputs and webpage after polling.
+            gamepad.tests.navigator.pollGamepads();
+            jqUnit.assertLeftHand("The gamepad should be connected with no buttons/axes disturbed initially.", gamepad.tests.modelAtRest, gamepad.tests.navigator.model);
+            jqUnit.assertEquals("The focus should not have changed.", document.querySelector("button"), document.activeElement);
```
the-t-in-rtf, line 73: I don't think you need to duplicate this, it's necessary for and already covered by the focus (tabbing) tests.

tests/js/click/click-mock-utils.js
```diff
@@ -0,0 +1,73 @@
...
+        // Initial count is set to 2 in all cases.
+        if (gamepad.tests.count === 1 || gamepad.tests.count === 3) {
+            inputSpec.buttons[inputButton] = 1;
+        }
+
+        // Return a list of gamepad objects (with no buttons/axes disturbed by default).
+        gamepadMock = gamepad.tests.utils.generateGamepadMock(inputSpec);
+
+        // Reduce the count on every call.
+        gamepad.tests.count--;
```
the-t-in-rtf, line 44: Manipulating a global variable to track the click state seems dangerous, especially if you have multiple tests in the same file.  I also don't quite understand the pattern, i.e. why would it ever be 3?

I understand not wanting to make a component just to track the state, one idea might be to stash the number of times an element has been clicked as an attribute of the element itself.  That would help you avoid leaving key code inline in an onclick attribute as well.

tests/html/click-tests.html
```diff
@@ -0,0 +1,82 @@
...
+            <br>
+            <button type="button" onclick="gamepad.tests.clicked = true;">Sample Button</button>
```
the-t-in-rtf, line 33: As I mentioned elsewhere, I'm not as big a fan of using global variables for this, especially when the code is inlined as part of the markup and easy to miss.  I would suggest adding an onClick listener to all elements that support a click that stashes the number of times clicked in an attribute like `timesClicked`.  That would give you a very easy way to look for changes without having to write inline code like this.

tests/js/click/click-basic-tests.js
```diff
@@ -0,0 +1,143 @@
...
+        fluid.registerNamespace("gamepad");
+        fluid.registerNamespace("gamepad.tests");
+
+        jqUnit.module("Gamepad Navigator Basic Click Tests", {
+            setup: function () {
+                gamepad.tests.windowObject = window;
+                gamepad.tests.count = 2;
+
+                // To test the gamepad's initial state.
+                gamepad.tests.modelAtRest = {
```
the-t-in-rtf, line 29: You should make a common helper function to take care of this and not repeat it in every new fixture.

tests/js/click/click-basic-tests.js
```diff
@@ -0,0 +1,143 @@
...
+                    connected: true,
+                    axes: {},
+                    buttons: {}
+                };
+
+                // Initialize in accordance with the 18 buttons on the PS4 controller.
```
the-t-in-rtf, line 35: Same here.

tests/js/click/click-basic-tests.js
```diff
@@ -0,0 +1,143 @@
...
+                for (var buttonNumber = 0; buttonNumber < 18; buttonNumber++) {
+                    gamepad.tests.modelAtRest.buttons[buttonNumber] = 0;
+                }
+
+                // Initialize in accordance with the 4 axes on the PS4 controller.
+                for (var axesNumber = 0; axesNumber < 4; axesNumber++) {
```
the-t-in-rtf, line 41: Same here.

tests/js/click/click-basic-tests.js
```diff
@@ -0,0 +1,143 @@
...
+            // Confirm that the instance of the gamepad navigator is created.
+            gamepad.tests.navigator = gamepad.inputMapper({
+                windowObject: gamepad.tests.windowObject
+            });
+            jqUnit.assertTrue("The Gamepad Navigator should be instantiated.", fluid.isComponent(gamepad.tests.navigator));
+
+            // Check the state of gamepad inputs and webpage after polling.
+            gamepad.tests.navigator.pollGamepads();
+            jqUnit.assertLeftHand("The gamepad should be connected with no buttons/axes disturbed initially.", gamepad.tests.modelAtRest, gamepad.tests.navigator.model);
+            jqUnit.assertEquals("The focus should not have changed.", document.querySelector("#radio-one"), document.activeElement);
```
the-t-in-rtf, line 102: IMO the focus check should be removed throughout these tests.

tests/js/click/click-basic-tests.js
```diff
@@ -0,0 +1,143 @@
...
+
+            // Update the gamepad to click on the radio button.
+            gamepad.tests.navigator.pollGamepads();
+
+            // Check if the radio button has been clicked.
+            jqUnit.assertTrue("The radio button should have been selected after clicking.", document.querySelector("#radio-one").checked);
+
+            // Restore the gamepad back to its neutral state i.e., release the button.
```
the-t-in-rtf, line 110: I'm a little confused about how/whether the button is released by this call, you didn't change the list of gamepads.  Is this what the count variable you used does, limit the activity to one polling cycle?

tests/js/click/click-dropdown-tests.js
```diff
@@ -0,0 +1,126 @@
...
+                gamepad.tests.count = 4;
+
+                // To test the gamepad's initial state.
+                gamepad.tests.modelAtRest = {
+                    connected: true,
+                    axes: {},
+                    buttons: {}
+                };
+
+                // Initialize in accordance with the 18 buttons on the PS4 controller.
```
the-t-in-rtf, line 35: Again, all of this initial setup should be a helper function.

the-t-in-rtf: By "all" I mean lines 29-43.

the-t-in-rtf:
I tried this out, it's looking quite good.  Most of my comments are around tidying up the tests rather than the functionality itself.

tests/html/click-tests.html
dmahajan980: I wasn't aware of the fact that QUnit has a variation in regard to the types of divs it uses, so I haven't tried that before. Does it have a special use case in the tests? I'll try to go through it as it's something I haven't heard or read before.

tests/js/click/click-basic-tests.js
dmahajan980: I had kept it as a verifying point because the tests can also fail due to the fact that the radio, checkbox, button, or any other element isn't focused before we perform the action. But if you still think that we should remove it then please let me know.

tests/js/click/click-mock-utils.js
dmahajan980: The count variable is not the number of times the button or any other element is clicked. But rather to track at which call to the polling function, we have to update the gamepad (like pressing the button at the second call to the polling function and releasing it in the next call).

dmahajan980: Sorry, if you are confused by the comment at [line #35](https://github.com/fluid-lab/gamepad-navigator/pull/16/files/fa1f864f9c40f02ca58b936f3c9062baf4da8925#diff-613bbd0f1614a81ad43f0c869a97f8b4R35) then please ignore it. It's incorrect and I'll fix it in my next commit.

tests/js/click/click-basic-tests.js
dmahajan980: Each value of count is associated with one polling cycle, i.e., for each polling cycle, we have a new value of count that we changed (decreased) at the end of the previous polling cycle. So we set the value of the count variable before every test unit and then we know that in a particular polling cycle, we want the button to be pressed or be released (by using the value of count which we change in each polling cycle).

tests/html/click-tests.html
the-t-in-rtf: It's not well documented in the QUnit docs, but there is [a brief bit on it in ours](https://docs.fluidproject.org/infusion/development/jqUnit.html#testing-in-the-browser).  The only difference is a) the markup is not displayed and b) the markup is reset between runs. Like I said, I was just curious, I'd be happy with horizontal rule between the QUnit markup and yours just to distinguish.

tests/js/click/click-basic-tests.js
the-t-in-rtf: I see your point, although I dislike that it's literally copied and pasted so many times.  If you can make a helper function for this initial check, that would be better.

tests/js/click/click-mock-utils.js
the-t-in-rtf: Yes, I kind of figured that out.  I see why you had to do this, but at a minimum the count needs to be renamed and documented better.  I still don't like making it a global variable, though.  I know you have setup and teardown, but each gamepad is manipulating the same variable.  If you could make it a member variable of the gamepad itself somehow, that would be better.

tests/js/click/click-basic-tests.js
the-t-in-rtf: As we discussed, simply making it a helper function would be fine.

tests/js/click/click-basic-tests.js
the-t-in-rtf: Yes, I get that now.  I still think you can at least clean this up to have a variable that is local to the gamepad mock.

tests/html/click-tests.html
dmahajan980: Thank you for sharing this. I'll check it out.

tests/js/click/click-basic-tests.js
dmahajan980: That's a good suggestion and I think we'd have to do the same for other tests we wrote earlier. I'll address this for other tests in a separate pull, just to keep things simple for you to review.

dmahajan980:
@the-t-in-rtf I've made the changes as you suggested. Kindly review it again.

Thanks

src/js/content_scripts/input-mapper-utils.js
```diff
@@ -375,17 +383,29 @@ https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
 
                 // Toggle the SELECT dropdown.
                 if (!document.activeElement.getAttribute("size") || document.activeElement.getAttribute("size") === "1") {
+                    /**
+                     * Store the initial size of the dropdown in a separate attribute
+                     * (if specified already).
+                     */
+                    var initialSizeString = document.activeElement.getAttribute("size");
+                    if (initialSizeString) {
+                        document.activeElement.setAttribute("initialSize", parseInt(initialSizeString));
```
the-t-in-rtf, line 392: Great.

tests/js/click/click-basic-tests.js
```diff
@@ -113,22 +90,18 @@ https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
...
             };
 
             // Initialize the webpage, i.e., focus on the checkbox element.
             $("#checkbox-one").focus();
 
             // Confirm that the instance of the gamepad navigator is created.
             gamepad.tests.navigator = gamepad.inputMapper({
-                windowObject: gamepad.tests.windowObject
+                windowObject: gamepad.tests.windowObject,
+                members: { count: 2 }
```
the-t-in-rtf, line 102: Rather than pass this as option, I'd suggest defining a gamepad.tests.inputMapper grade that has the members and windowObject setting (if possible).

dmahajan980:
@the-t-in-rtf I've updated the files as you suggested. Kindly review it again.

Thank you :)

PR #19, opened 2020-07-08
Added feature for forward and backward history navigation

dmahajan980:
With the history navigation, it ensures that the last focused element (before the user moved forward/backward in history) gets focused when the user navigates back to the same page (through history navigation).


dmahajan980:
Judging from the errors we are getting, I think these are possibly due to the fact that the webpage cannot use the chrome extension APIs, as the pattern we are using in our tests involves adding the component files into the test window.

dmahajan980:
@the-t-in-rtf kindly have a look over this pull. If everything's fine then we can proceed with merging it.

the-t-in-rtf:
I did a bit more extensive testing before looking at the code, and will file a few tickets for the key issues I'm seeing.   Specific to this pull, I wasn't ever able to get the history navigation working with a gamepad.  I assume that's working for you?

the-t-in-rtf:
Disregard, I had to run `grunt build` and update the plugin rather than simply reload it.  IMO that should be added as a postinstall script to your package.json.

the-t-in-rtf:
OK, so I can get the back button to work, but I've never seen the forward button work.  I also can get into a state where it never stops trying to go back, i.e. if I use the mouse to navigate forward, it navigates right back.

the-t-in-rtf:
If I turn the extension off and navigate around, and then turn it back on, it navigates all the way back to the beginning of history.

the-t-in-rtf:
OK, with my PS4 controller disconnected I was able to get further.  I can get forward and backward navigation to work, but was able to get forward navigation "stuck" where it would keep going forwards long after the button was released.

the-t-in-rtf:
With the problems with "stuck" controls it's hard to tell, but it seemed like the focus wasn't being preserved.

the-t-in-rtf:
Seems like I had a stuck button on my controller.  I can at least verify that the forward/back controls work both with buttons and analog controls.

the-t-in-rtf:
I think the focus is working, although when it does work it seems to work very slowly.

dmahajan980:
@the-t-in-rtf the focus is restored after the webpage has finished loading. So, as you said, with the stuck forward/backward navigation, you might be navigating to the other pages in the history before the current web page gets loaded completely and because of that, the last focused element wasn't actually focused.

src/js/content_scripts/input-mapper-instance.js
```diff
@@ -10,9 +10,61 @@ You may obtain a copy of the BSD 3-Clause License at
...
+    gamepad.inputMapper({
+        /**
+         * To avoid errors that are generated due to the presence of Chrome Extension
+         * APIs.
+         */
+        listeners: {
+            "onCreate.restoreFocusBeforeHistoryNavigation": "{that}.restoreFocusBeforeHistoryNavigation"
+        },
+        invokers: {
+            restoreFocusBeforeHistoryNavigation: {
```
the-t-in-rtf, line 64: I think this can just be `restoreFocus` unless you're using it at multiple points.

src/js/content_scripts/input-mapper-instance.js
```diff
@@ -10,9 +10,61 @@ You may obtain a copy of the BSD 3-Clause License at
...
+        });
+    };
+
+    // Create an instance of the component.
+    gamepad.inputMapper({
+        /**
+         * To avoid errors that are generated due to the presence of Chrome Extension
+         * APIs.
+         */
+        listeners: {
```
the-t-in-rtf, line 60: Why isn't this in the defaults for `gamepad.inputMapper` instead?

src/js/content_scripts/input-mapper-utils.js
```diff
@@ -414,4 +406,94 @@ https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
...
+     * @param {Integer} value - The value of the gamepad input.
+     * @param {Boolean} invert - Whether the history navigation should be in opposite
+     *                           order.
+     *
+     */
+    gamepad.inputMapperUtils.thumbstickHistoryNavigation = function (that, value, invert) {
+        // Get the updated input value according to the configuration.
+        var inversionFactor = invert ? -1 : 1;
+        value = value * inversionFactor;
+        if (value > 0) {
```
the-t-in-rtf, line 426: You should really add a cutoff here instead of using zero.  I can get the thumbsticks on my Xbox One X controller to a resting state that's plus or minus `0.08` even.  My PS4 controller jitters constantly around its resting value, for example moving between 0.01 and 0.02 in whatever direction the thumbstick comes to rest in.

src/js/content_scripts/input-mapper.js
```diff
@@ -153,7 +153,7 @@ https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
                     },
                     // Right thumbstick vertical axis.
```
the-t-in-rtf, line 154: Given the way the controls appear on screen (back arrow, forward arrow), I'd argue that you should use the right thumbstick horizontal axis for this by default.

src/js/content_scripts/input-mapper-instance.js
dmahajan980: I placed it in the defaults initially in my first commit. But since they use the `chrome.storage` API, our tests fail as the browser is not aware of the Chrome Extension APIs. So to avoid those test failures, I moved these here in my second commit.

src/js/content_scripts/input-mapper-utils.js
dmahajan980: We don't need to add the cutoff value here as even if these bypass the `0` value condition, they'll have to go through the `cutoffValue` when the `nextPageInHistory` ([here](https://github.com/fluid-lab/gamepad-navigator/pull/19/files/27188c6358e1b5c2dea7346eaee4120ec60b7bf5#diff-cc20e237566e63150ac1d82f0dbee3f8R476)) or `previousPageInHistory` ([here](https://github.com/fluid-lab/gamepad-navigator/pull/19/files/27188c6358e1b5c2dea7346eaee4120ec60b7bf5#diff-cc20e237566e63150ac1d82f0dbee3f8R443)) is called. So, in your case, if the value is plus/minus `0.08`, it would manage to call the `nextPageInHistory` or `previousPageInHistory` method, but the `cutoffValue` inside these methods will prevent anything from happening.

What I simply mean to say is we do have the `cutoffValue` as the check, but inside the method it calls. However, I can replace `0` with `cutoffValue` here as well if you want to keep it that way.

src/js/content_scripts/input-mapper-instance.js
the-t-in-rtf: I see.  I would still rather see them as defaults somewhere, so that people can more easily / safely extend your work.  As it stands, anyone wanting to extend your work would have to copy this into their own.  I would suggest abstracting out a base grade such as `gamepad.inputMapper.base` with everything that won't break the tests, and then having a smaller `gamepad.inputMapper` grade that includes this.

src/js/content_scripts/input-mapper-utils.js
the-t-in-rtf: I guess it's fine as long as the net effect is that jitter is screened out.  Down the road we may need to talk about doing this further upstream, i.e. protecting everything that is based on jittery controls at the point where we read it from the gamepad rather than requiring multiple cutoff checks downstream.  I'm happy to leave it for now, I'd rather you focus on the tab stability issues for now.

dmahajan980:
@the-t-in-rtf Thanks for your feedback :)
I've made the changes as you suggested. Kindly review it when you have time.

PR #21, opened 2020-07-10
Added feature to open and close tabs

dmahajan980:
For the above features, I added background scripts, which are another component of chrome extension to handle browser level activities (such as modify tabs, windows, et cetera). This pull also includes minor optimizations to remove some of the existing redundancy and for more readability.

### DIFF
--- a/src/js/content_scripts/input-mapper-base.js
+++ b/src/js/content_scripts/input-mapper-base.js
@@ -29,7 +29,7 @@ https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
                         defaultAction: "click",
                         currentAction: null,
                         speedFactor: 1,
-                        background: true
+                        background: false
                     },
                     // Face Button.
                     // Circle on PlayStation controller & B on Xbox controller.
@@ -37,7 +37,7 @@ https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
                         defaultAction: null,
                         currentAction: null,
                         speedFactor: 1,
-                        background: true
+                        background: false
                     },
                     // Face Button.
                     // Square on PlayStation controller & X on Xbox controller.
@@ -45,7 +45,7 @@ https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
                         defaultAction: "previousPageInHistory",
                         currentAction: null,
                         speedFactor: 1,
-                        background: true
+                        background: false
                     },
                     // Face Button.
                     // Triangle on PlayStation controller & Y on Xbox controller.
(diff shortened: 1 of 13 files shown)


dmahajan980:
@the-t-in-rtf I noticed that the CI fails in the first run. This happened in the previous pull as well, and this is the error message it logs:

```bash
Error: Browser failed to connect within 30s. testem.js not loaded?
```

For more information, here's the [log file](https://pipelines.actions.githubusercontent.com/tGIr5uyvF7VN3ezv2bRkYx7iMDAcO7cKG7DOPflSRanmeS4u39/_apis/pipelines/1/runs/9/signedlogcontent/3?urlExpires=2020-07-10T21%3A00%3A10.1745618Z&urlSigningMethod=HMACV1&urlSignature=WWh084uKgDySK3S6h2uPCByYkJVm9kRVzqzCmdYRQSA%3D) for the run.

the-t-in-rtf:
I just manually tested this with two controllers and noticed a few things that need work before we can merge this:

1. I can't add a tab from the "New Tab" page.
2. I can't add a tab from the "extensions" page, either.
3.  If I open a page other than the "New Tab" or "extensions" page (such as Wikipedia), I can create new tabs by hitting the start/options button.
4. Although I can close "real" tabs, I cannot close any instances of the "New Tab" or "extensions" page.  In other words, I cannot close the tabs I myself opened.

There are also a few things that you need to document and work on when you move on to other related features:

1.  If I create a new tab, the focus does not change, i.e. the tab that was just added should receive focus.  You may need to address this once you add a means of changing browser focus, i.e. browser tab navigation.  A TODO would be sufficient for now.
2. I don't seem to be able to add a tab if there are no open windows.  When you add support for adding a new window, you should also ensure that a new window opens if someone tries to create a tab when no windows are open.  A TODO is sufficient for now.
3. I would suggest adding a feature to reopen the last closed tab as a separate issue.  I use this often myself, and it is useful if someone accidentally closes an important tab (or window).

the-t-in-rtf:
> @the-t-in-rtf I noticed that the CI fails in the first run. This happened in the previous pull as well, and this is the error message that fails the run:
> 
> ```shell
> Error: Browser failed to connect within 30s. testem.js not loaded?
> ```

The link to the log file expired.  The CI run associated with the commit did seem to work.  If you see it happen again in a build triggered by a commit, send me a link to the commit and I can look at it.

src/js/content_scripts/input-mapper-base.js
```diff
@@ -28,100 +28,116 @@ https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
...
+                        background: true
                     },
                     // Select/Share on PlayStation controller & Back on Xbox controller.
                     "8": {
-                        defaultAction: null,
+                        defaultAction: "closeCurrentTab",
                         currentAction: null,
-                        speedFactor: 1
+                        speedFactor: 1,
+                        background: true
```
the-t-in-rtf, line 91: This structure is getting complex enough that it's probably time to document what elements are supported and what each means.  I'll create a separate ticket with examples of how we've handled this in the past.

src/js/content_scripts/input-mapper-base.js
```diff
@@ -288,19 +304,18 @@ https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
         var inputProperties = that.model.map[inputType][index],
             actionLabel = fluid.get(inputProperties, "currentAction") || fluid.get(inputProperties, "defaultAction");
 
+        /**
+         * TODO: Modify the action call in such a manner that the action gets triggered
+         * when the inputs are released. (To gain shortpress and longpress actions)
```
the-t-in-rtf, line 309: Once you've got the core functionality on board in some form (performing an action using at least one type of input), we should review and refactor and make each "style" of interaction into a clearer pattern and ensure that all functions can be used with:

1. Short presses that trigger an action on button release.
2. Long presses that trigger an action on button release after a configured time period.
3. Long presses that repeatedly trigger an action.

You'll need a kind of matrix for these, where you support each type of action for digital buttons, analogue buttons, and axes.  This is probably one to handle later, for example, after you've got the configuration panel working.

dmahajan980:
> 1. I can't add a tab from the "New Tab" page.
> 2. I can't add a tab from the "extensions" page, either.
> 3. If I open a page other than the "New Tab" or "extensions" page (such as Wikipedia), I can create new tabs by hitting the start/options button.
> 4. Although I can close "real" tabs, I cannot close any instances of the "New Tab" or "extensions" page. In other words, I cannot close the tabs I myself opened.

@the-t-in-rtf this is because of the fact that no Chrome extension can run on these pages (New Tab, Downloads, Extensions). 
Chrome and other Chromium-based browsers forbid running of extension on these pages and so the gamepad can't be used there.
(_More information can be found [here](https://codereview.chromium.org/2978953002/)_)

One thing we can do is to make the tab open with an address/webpage (such as Google), which would allow the user to navigate further. This would also help in switching tabs feature, as then we won't have any barrier that'll stop switching the tabs seamlessly.

> If I create a new tab, the focus does not change, i.e. the tab that was just added should receive focus. You may need to address this once you add a means of changing browser focus, i.e. browser tab navigation. A TODO would be sufficient for now.

This is simply a configuration option, i.e., the user can choose this mode or the other mode of tabs to be focused when opened. The [background option](https://github.com/fluid-lab/gamepad-navigator/pull/21/files#diff-b1b19b4b4789c274fa2633af3eb03617R32) in the configuration maps handles the way these tabs open. So if you set the background option to false, the tab will receive focus when opened.

> I don't seem to be able to add a tab if there are no open windows. When you add support for adding a new window, you should also ensure that a new window opens if someone tries to create a tab when no windows are open. A TODO is sufficient for now.

Again, this is due to the fact that the Chrome extension doesn't operate on the New Tab and other pages native to the Chrome Browser. The user has to navigate to some webpage first to activate their gamepads and then navigate using the gamepad.

> I would suggest adding a feature to reopen the last closed tab as a separate issue. I use this often myself, and it is useful if someone accidentally closes an important tab (or window).

That's a really good suggestion and I think this is something people frequently use. I'll address this in a separate pull.

the-t-in-rtf:
> One thing we can do is to make the tab open with an address/webpage (such as https://www.google.com/), which would allow the user to navigate further. This would also help in switching tabs feature, as then we won't have any barrier that'll stop switching the tabs seamlessly.

This seems the most reasonable option, but do please make the URL configurable, and add a note to the README about the fact that we have to override the "new tab" screen.

the-t-in-rtf:
> The [background option](https://github.com/fluid-lab/gamepad-navigator/pull/21/files#diff-b1b19b4b4789c274fa2633af3eb03617R32) in the configuration maps handles the way these tabs open

Once you have integration tests, you would need to test both, for now I would argue that setting it to `false` makes more sense.  Once you have browser tab switching it will matter less, but focusing on the tab is the default behaviour when using the keyboard bindings, and I'd argue we should mimic that.

src/js/content_scripts/input-mapper-base.js
dmahajan980: Yes, I think a separate ticket would be sufficient for the purpose.

dmahajan980:
@the-t-in-rtf thanks for the feedback. I've made the changes as you suggested. Kindly review it.

PR #25, opened 2020-07-14
Added features for forward and backward browser tab navigation using buttons

dmahajan980:


the-t-in-rtf:
Just finished the manual QA, looks good.  I'll review the code now.

src/js/background_scripts/message-listener-utils.js
```diff
@@ -28,4 +28,53 @@ https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
     gamepad.messageListenerUtils.openNewTab = function (active, homepageURL) {
         chrome.tabs.create({ active: active, url: homepageURL });
     };
+
+    /**
+     *
+     * Switch to the previous tab in the available order in the current window.
```
the-t-in-rtf, line 34: You can leave out "in the available order" here.

src/js/background_scripts/message-listener-utils.js
```diff
@@ -28,4 +28,53 @@ https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
...
+            }
+        });
+    };
+
+    /**
+     *
+     * Switch to the next tab in the available order in the current window.
+     *
+     */
+    gamepad.messageListenerUtils.goToNextTab = function () {
```
the-t-in-rtf, line 63: These should not be near duplicates.  I would suggest a single function with a direction, so that you're not repeating the "number of tabs" checks and other code.

src/js/content_scripts/input-mapper-base.js
```diff
@@ -221,6 +221,11 @@ https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
                 funcName: "gamepad.inputMapper.base.clearIntervalRecords",
                 args: ["{that}.intervalRecords"]
             },
+            // TODO: Investigate, identify, and fix tab navigation issues.
```
the-t-in-rtf, line 224: Such as?  Do you already know of any issues?

src/js/content_scripts/input-mapper.js
```diff
@@ -28,6 +28,14 @@ https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
                 funcName: "gamepad.inputMapper.restoreFocus",
                 args: ["{that}.options.windowObject", "{that}.tabindexSortFilter"]
             },
+            goToPreviousTab: {
```
the-t-in-rtf, line 31: Since you already have invokers, if you were to standardise to a single function, you could just indicate the direction here.

src/js/content_scripts/input-mapper-base.js
dmahajan980: These are a reference to the tab navigation issues you faced and not the browser tab navigation. There's nothing new here, just rearranged the TODO along with the invoker.

dmahajan980:
@the-t-in-rtf thanks for the feedback. I've made the changes as you suggested. Could you please take a look over it again and suggest if there's anything else I need to change or add?

src/js/content_scripts/input-mapper-base.js
the-t-in-rtf: Got it.

PR #31, opened 2020-07-14
Added gamepad.navigator documentation

dmahajan980:
The documentation follows the same pattern as used in [Fluid Express Docs](https://github.com/fluid-project/fluid-express/tree/master/docs).

### DIFF
--- a/docs/components/navigator.md
+++ b/docs/components/navigator.md
@@ -1,0 +1,115 @@
+<!--
+Copyright (c) 2020 The Gamepad Navigator Authors
+See the AUTHORS.md file at the top-level directory of this distribution and at
+https://github.com/fluid-lab/gamepad-navigator/raw/master/AUTHORS.md.
+
+Licensed under the BSD 3-Clause License. You may not use this file except in
+compliance with this License.
+
+You may obtain a copy of the BSD 3-Clause License at
+https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
+-->
+
+# gamepad.navigator
+
+This component listens to the `gamepadconnected` and `gamepaddisconnected` events in the browser. When at least one
+gamepad is connected, it reads inputs from them at a particular frequency and stores those input values in the
+component's model data. The component stops reading these inputs when the last gamepad is disconnected.
+
+## Using this grade
+
+This grade can work without passing any option to it. An instance of the `navigator` does not require any options to be
+passed and can listen to the gamepad events and read the gamepad inputs. To use it, we can create an instance of the
+component or use it as a grade another component:
+
+``` javascript
+// Either create an instance of the navigator.
+var navigatorInstanceOne = gamepad.navigator();
+
+// Otherwise create a custom component using the navigator grade.
+fluid.defaults("my.navigator.grade", {
+    gradeNames: ["gamepad.navigator"]
+};
+var navigatorInstanceTwo = my.navigator.grade();
+```
+
+## Component Options
+
+The following component configuration options are supported:
+
+| Option | Type | Default Value | Description |
+| :---: | :---: | :---: | :--- |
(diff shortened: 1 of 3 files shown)


docs/components/navigator.md
```diff
@@ -0,0 +1,115 @@
...
+https://github.com/fluid-lab/gamepad-navigator/raw/master/AUTHORS.md.
+
+Licensed under the BSD 3-Clause License. You may not use this file except in
+compliance with this License.
+
+You may obtain a copy of the BSD 3-Clause License at
+https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
+-->
+
+# gamepad.navigator
```
the-t-in-rtf, line 13: It's a small thing, but I prefer to enclose component names in backticks, i.e. `gamepad.navigator`.

docs/components/navigator.md
```diff
@@ -0,0 +1,115 @@
...
+
+This component listens to the `gamepadconnected` and `gamepaddisconnected` events in the browser. When at least one
+gamepad is connected, it reads inputs from them at a particular frequency and stores those input values in the
+component's model data. The component stops reading these inputs when the last gamepad is disconnected.
+
+## Using this grade
+
+This grade can work without passing any option to it. An instance of the `navigator` does not require any options to be
```
the-t-in-rtf, line 21: The first and second sentences are somewhat redundant.  I would change both to something like "You can change the configuration options by either extending the component or by passing your own options.  If no options are passed, the defaults are used."

docs/components/navigator.md
```diff
@@ -0,0 +1,115 @@
...
+fluid.defaults("my.navigator.grade", {
+    gradeNames: ["gamepad.navigator"],
+    windowObject: myCustomWindowObject,
+    frequency: 100
+};
+var navigatorInstanceTwo = my.navigator.grade();
+```
+
+Though the `windowObject` is a configurable option, it should have all the properties and methods that the navigator
+uses. Otherwise, the navigator won't work and will throw errors. You can also use a mutated `window` object and pass it
```
the-t-in-rtf, line 64: "modified" is probably better than "mutated", and I would give an example, such as "a modified window object, such as the mocks used in this package's tests)"

docs/components/navigator.md
```diff
@@ -0,0 +1,115 @@
...
+these events are fired, the navigator fires its native `onGamepadConnected` or `onGamepadDisconnected` events,
+depending upon the former event. These events have listeners attached to it, which start reading the gamepad inputs or
+stop reading it further.
+
+### `{navigator}.onConnected(that)`
+
+- `that {Object}` The gamepad navigator component.
+- Returns: Nothing.
+
+It is the listener for the gamepad navigator component's event `onGamepadConnected`. When it is called, it initiates
```
the-t-in-rtf, line 85: You can simplify all of these "it is the listener" phrases to simply say "Listens".

docs/components/navigator.md
```diff
@@ -0,0 +1,115 @@
...
+the gamepad polling invoker [`{navigator}.pollGamepads`](#navigatorpollgamepadsthat) to be called at the same frequency
+as specified in the component's configurable option `frequency`. The interval ID is stored in the component's member
+`connectivityIntervalReference` for reference in other invokers and listeners present in the component.
+
+### `{navigator}.pollGamepads(that)`
+
+- `that {Object}` The gamepad navigator component.
+- Returns: Nothing.
+
+It is called in intervals for reading the gamepad inputs and updating the navigator component's model with those
```
the-t-in-rtf, line 95: It's a style thing, but I dislike leading with "it" if we can help it.  You can simplify this to "Called periodically to read the gamepad...."

the-t-in-rtf:
The level of detail here is fine, I left a few small comments.  When you get to the input mapper, I think you will definitely need to use diagrams or a table to illustrate the range of buttons and the default mappings.

PR #32, opened 2020-07-14
Fixed tabbing issues and changed button tab navigation to single-press mode

### LINKED ISSUE #14: Multiple button presses required to tab forwards or backwards
With Chrome on OS X, I have to press the bumpers multiple times (sometimes many times) to get the focus to move.  This happens with both the PS4 and Xbox One X controller, and on various sites including [Wikipedia](https://wikipedia.com/).

dmahajan980:
Fixes #14


the-t-in-rtf:
Very happy with the code, especially refactoring to use a common function.  The "incidental" tabbing is really rock solid, this will greatly improve the upcoming demos.

PR #33, opened 2020-07-15
Added feature to open and close windows

### LINKED ISSUE #26: Add support for opening and closing windows.
As stated in the proposal, a key feature of the gamepad navigator is the ability to open and close windows.

dmahajan980:
Resolves #26

### DIFF
--- a/src/js/background_scripts/message-listener-utils.js
+++ b/src/js/background_scripts/message-listener-utils.js
@@ -66,4 +66,81 @@ https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
             }
         });
     };
+
+    /**
+     *
+     * Open a new window.
+     *
+     * @param {Boolean} active - Whether the new window should be focused when created.
+     * @param {String} homepageURL - The URL for the new window.
+     *
+     */
+    gamepad.messageListenerUtils.openNewWindow = function (active, homepageURL) {
+        chrome.windows.getAll(function (windowsArray) {
+            /**
+             * Minimize the current window to avoid gamepad action execution in multiple
+             * windows (if new window is not to be opened in background). Also minimize
+             * all other windows previously opened before connecting the gamepad.
+             */
+            chrome.windows.getCurrent(function (currentWindow) {
+                windowsArray.forEach(function (window) {
+                    var windowUpdate = { state: "minimized" };
+
+                    /**
+                     * Skip minimization of current window if the new window is to be
+                     * opened in the background.
+                     */
+                    if (window === currentWindow && !active) {
+                        return;
+                    }
+
+                    // Minimize the window.
+                    chrome.windows.update(window.id, windowUpdate);
+                });
+            });
+
+            // Open a new window (in active mode or background).
+            chrome.windows.create({
+                focused: active,
+                url: homepageURL,
+                state: "maximized"
+            });
+        });
+    };
+
+    /**
+     *
+     * Close the currently opened window.
+     *
+     */
+    gamepad.messageListenerUtils.closeCurrentWindow = function () {
+        chrome.windows.getCurrent(function (currentWindow) {
+            chrome.windows.getAll(function (windowsArray) {
+                // Remove the currently active window object from the windows array.
+                windowsArray = windowsArray.filter(function (window) {
+                    return window.id !== currentWindow.id;
+                });
+
+                /**
+                 * Minimize all windows other than the current window before connecting
+                 * the gamepad.
+                 */
+                windowsArray.forEach(function (window) {
+                    var windowUpdate = { state: "minimized" };
+                    chrome.windows.update(window.id, windowUpdate);
+                });
+
+                // Close the current window.
+                chrome.windows.remove(currentWindow.id);
+
+                // Activate / maximize the last active window.
+                if (windowsArray.length) {
+                    chrome.windows.update(windowsArray[windowsArray.length - 1].id, {
+                        state: "maximized",
+                        focused: true
+                    });
+                }
+            });
+        });
+    };
 })(fluid);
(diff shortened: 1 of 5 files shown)


the-t-in-rtf:
So, I tested this and have a few questions (some of which I suspect I know the answer to).

First, I see that this only works if you already have a "normal" window open?  Given that you can close the last open window, this may be a bit of a problem, as you can pretty easily get into a state where you can no longer do anything with the gamepad alone.

Second, I notice that the previous window is minimised whenever you open a new window.  Was this necessary in order to accomplish the goal?

dmahajan980:
@the-t-in-rtf I can understand the concern with your comment. So I'll try to answer your questions with the facts that I'm aware of.

> First, I see that this only works if you already have a "normal" window open? Given that you can close the last open window, this may be a bit of a problem, as you can pretty easily get into a state where you can no longer do anything with the gamepad alone.

When we have closed all the normal windows, I understand two scenarios:
1. All the windows have been closed, i.e., the browser itself is not running at the moment. In this case, I don't think there's any possibility of opening the browser window(s) using the gamepad. The browser (which runs the extension) is not running and couldn't read the gamepad inputs to take some action.
2. Another case could be when you have the browser running, but only the native windows (New Tab, Extensions Page, Downloads Page, et cetera) are open. In this case, we can't do anything with the gamepad as the Chrome extension cannot run on these pages (as discussed [here](https://github.com/fluid-lab/gamepad-navigator/pull/21#issuecomment-657460478) in the recent pull).

Other than the above cases, it should work everywhere. But if you meant something else, then please let me know.

> Second, I notice that the previous window is minimised whenever you open a new window. Was this necessary in order to accomplish the goal?

I am minimizing the windows because all the non-minimized windows will open a new window when you try to do so using the gamepad. When you perform the action using the gamepad, it won't open a single window, but rather double up the existing number of windows (as each non-minimized window will open a new window). I'm guessing that multiple active windows are reading the same gamepad input, and the net effect of the action on each active window adds up.

To prevent this from happening, I went with the approach of minimizing all windows. The current window won't minimize in cases where you chose to open a new window in the background. Although I've covered all the use cases I could think of, I guess we would later have to test it extensively.

the-t-in-rtf:
> I am minimizing the windows because all the non-minimized windows will open a new window when you try to do so using the gamepad.

This highlights a larger issue, i.e. all gamepad inputs are read for all windows.  I will write up a separate ticket for that and we can leave your approach in place for now.

src/js/background_scripts/message-listener-utils.js
```diff
@@ -66,4 +66,81 @@ https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
...
+    /**
+     *
+     * Open a new window.
+     *
+     * @param {Boolean} active - Whether the new window should be focused when created.
+     * @param {String} homepageURL - The URL for the new window.
+     *
+     */
+    gamepad.messageListenerUtils.openNewWindow = function (active, homepageURL) {
+        chrome.windows.getAll(function (windowsArray) {
```
the-t-in-rtf, line 79: Please leave a TODO here indicating that this should be removed once #34 is addressed.

src/js/content_scripts/input-mapper-background-utils.js
```diff
@@ -91,4 +91,41 @@ https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
...
+     * Sends message to the background script to open a new window.
+     *
+     * @param {Object} that - The inputMapper component.
+     * @param {Integer} value - The value of the gamepad input.
+     * @param {Boolean} background - Whether the new window should open in background.
+     * @param {Integer} oldInputValue - The previous value of the gamepad input.
+     * @param {String} homepageURL - The URL for the new window.
+     *
+     */
+    gamepad.inputMapperUtils.background.openNewWindow = function (that, value, background, oldInputValue, homepageURL) {
```
the-t-in-rtf, line 106: You now have a lot of functions like this where you screen for `value < oldInputValue` and then construct an action payload.  It seems like it's time to replace all six functions that call `chrome.runtime.sendMessage` with invokers that are backed by a common function.

This would also leave more of the key wiring declared as options, which I prefer to having so much of it "in code".

src/js/content_scripts/input-mapper.js
```diff
@@ -80,6 +88,30 @@ https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
...
-    gamepad.inputMapper();
+    // Create an instance of the inputMapper when a new page is opened.
+    gamepad.inputMapperInstance = null;
+    gamepad.inputMapperInstance = gamepad.inputMapper();
+
+    /**
+     * Restore or destroy the instance of the inputMapper when the visibility of current
+     * window/tab is changed.
+     */
+    document.addEventListener("visibilitychange", function () {
```
the-t-in-rtf, line 99: So, I don't understand why this doesn't address #34.  I mean, if you're managing to destroy the input mapper, how are the other windows able to take action on inputs?

the-t-in-rtf:
I have created tickets for a few key issues that IMO should probably be out of scope for this PR.  I also left you a few key things to fix/discuss.

src/js/content_scripts/input-mapper.js
dmahajan980: This handles most of the parts when the window is minimized, or the tab is changed (though we don't have problems in tabs). But when you open a new window, the old window is still running (as what I observed) and this [Visibility API](https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API) doesn't account for that. So, those issues are handled by it, but on minimizing the window.

dmahajan980:
@the-t-in-rtf I think I have made a solution that works on the currently active window/tab without using minimization and addresses #34. But I think it'd be better if we address that in a separate pull. So, if there isn't anything that we need to change here any further, we should merge this.

the-t-in-rtf:
That's awesome, I agree that's better as a separate pull.

Issue #34, opened 2020-07-16
Gamepad inputs should only affect the currently focused window.


the-t-in-rtf:
As discussed in #33, all non-minimised open windows attempt to act on gamepad input at the same time.  This is undesirable, and leads to issues like:

1. Scrolling or tab navigating in one window scrolls and tab navigates in all of them.
2. Attempting to click a link in the active window can result in unexpectedly clicking a link in another window.
3. Hitting the "new tab" button results in one new tab per active window (all created in the active window).
4. Hitting the "close tab" button results in one tab being closed per active window (in the active window).
5. Hitting the "new window" button results in one window being created per active window.
6. Hitting the "close window" button with multiple windows open results in a loop where new windows are continuously created and minimised.

We need to find a way to ensure that gamepad input is only processed once, and that action is taken only  on the currently focused window.  I notice that history navigation appears to work as expected, most other functions do not.

PR #36, opened 2020-07-17
Added feature to switch windows in forward and reverse direction and modified the structure to read gamepad inputs in current window/tab only

### LINKED ISSUE #27: Add ability to navigate between browser windows.
Once we add the ability to add/close windows, we need to add the ability to switch between open windows.

dmahajan980:
Resolves #27 and #34

### DIFF
--- a/src/js/background_scripts/message-listener-utils.js
+++ b/src/js/background_scripts/message-listener-utils.js
@@ -76,41 +76,10 @@ https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
      *
      */
     gamepad.messageListenerUtils.openNewWindow = function (active, homepageURL) {
-        chrome.windows.getAll(function (windowsArray) {
-
-            /**
-             * TODO: Remove the minimization part when the issue #34 is fixed, i.e., when
-             * only the current window will read gamepad inputs.
-             */
-
-            /**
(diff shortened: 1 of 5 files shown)


the-t-in-rtf:
The code in general looks solid.  I'm a little concerned about a seemingly new behaviour I'm seeing intermittently.  The tab navigation seems to be going forward or backwards by two "steps".  I got it to go away by reloading a particular page, but I saw it more than once.  I'll open a ticket if we see it more often.

src/js/background_scripts/message-listener-utils.js
```diff
@@ -121,32 +90,49 @@ https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
...
+     *
+     */
+    gamepad.messageListenerUtils.switchWindow = function (windowDirection) {
+        chrome.windows.getAll(function (windowsArray) {
+            // Switch only if more than one window is present.
+            if (windowsArray.length > 1) {
+                // Find index of the currently active window.
+                var focusedWindowIndex = null;
+                windowsArray.forEach(function (window, index) {
+                    if (window.focused) {
```
the-t-in-rtf, line 111: I would suggest using [`fluid.find`](https://docs.fluidproject.org/infusion/development/CoreAPI.html#fluidfindsource-func-deflt) here, that way the search will stop once you find the window with focus.

PR #37, opened 2020-07-23
Added configuration panel

dmahajan980:
This pull request lacks the list for switching between the configuration menus of different buttons, axes, et cetera, as it has become larger than I expected. Other than that, everything we discussed has been covered.


dmahajan980:
I think I need some more time to work on it as I'm starting to see some issues. I'll update the pull request as quickly as I can.

the-t-in-rtf:
Thanks.  When you're ready for review, I would appreciate a quick summary of what has and hasn't been done, i.e. what you will do in later pulls, so that I know how to evaluate what you have.

dmahajan980:
@the-t-in-rtf I've fixed the issues I noticed, but I'm still not sure if it's completely bug-free. Below is a list of things I've covered:

- Open the configuration panel using the keyboard shortcut `Ctrl + Alt + 5`.
- Added description for each gamepad input.
- Relabelled all the buttons (Save Changes, Set All to None, et cetera).
- Added the ability to display action-specific configuration options (such as `Speed Factor`, `Invert`, et cetera) accordingly.
- Configuration Panel now displays according to the color mode of the browser (Light Mode and Dark Mode).
- `Save Changes` button only works when the current configuration is changed (turns Green from Grey).
- Added the `Set All to None` button to neutralize all the inputs.

The only thing we need to cover is to add a list for switching between different inputs' configuration menu.

the-t-in-rtf:
OK, I have tested this manually and confirmed that:

1. The "Set All to None" button works.
2. The "Restore Defaults" button works.
3. You can set controls for individual buttons as expected.
4. All windows use the new control schema when you save your changes.
5. Changes are saved across restarts.

I will look at the code as well, but here's some feedback just based on using it.  First, we discussed removing the progress bar, which I see is still there. I would make a note to remove or update that when you implement the "list view".

More importantly, if the panel closes, you immediately lose the unsaved state.  This is particularly a problem as tabbing past the last element in the panel closes the panel.  So, if you tab past "Save Changes" accidentally, you can easily lose all the work you've done tabbing through all the options.  Ideally you should warn the users that they're about to lose state or preserve the unsaved state somehow.

This suggests another thing to me.  What if the user is making changes and actually wants to abandon them, i.e. go back to their previous setting?  I guess closing the dialogue will do this, but IMO there should be an explicit "cancel" button.  IMO you have room for a "Discard Changes" button to the left of "Save Changes", which might just close the dialogue, but. you'd still have the problem of a stray click also resulting in a loss of state.

Are there options to persist the unsaved state temporarily?   Keeping it for the life of the browser session is probably adequate.

dmahajan980:
@the-t-in-rtf thanks for testing this out.

> First, we discussed removing the progress bar, which I see is still there. I would make a note to remove or update that when you implement the "list view."

I am surprised that you see the progress bar as I removed it in my previous commit, and it doesn't show up here (tried on Ubuntu and Windows 10). But I'll recheck it.

> More importantly, if the panel closes, you immediately lose the unsaved state. This is particularly a problem as tabbing past the last element in the panel closes the panel. So, if you tab past "Save Changes" accidentally, you can easily lose all the work you've done tabbing through all the options. Ideally you should warn the users that they're about to lose state or preserve the unsaved state somehow.

I think it's a good idea to show some warning to users that they're closing the panel without saving changes, and I'll address it in my next commit. By the way, I'm not sure if the panel closes in your system once you tab past the "Save Changes" button as it tabs to the first configuration menu on Ubuntu and Windows 10. Could you please confirm if that's the case?

> IMO you have room for a "Discard Changes" button to the left of "Save Changes", which might just close the dialogue,

That's a good suggestion. But I need to check if it's feasible to close the panel programmatically. Otherwise, we can have this button just to restore the latest saved configuration, at least.

> Are there options to persist the unsaved state temporarily?

I don't think so, as the Chrome Extension pop-ups (configuration panel) is cleared when closed. But I'll try searching a bit to find anything that works. Maybe I can create my own mechanism with Chrome's Storage API. Whatever the case it would be, I'll update you.

the-t-in-rtf:
> I am surprised that you see the progress bar as I removed it in my previous commit, and it doesn't show up here (tried on Ubuntu and Windows 10). But I'll recheck it.

Sorry, I had to run `grunt build` and was misremembering that part.  That's fixed.

the-t-in-rtf:
> By the way, I'm not sure if the panel closes in your system once you tab past the "Save Changes" button as it tabs to the first configuration menu on Ubuntu and Windows 10. Could you please confirm if that's the case?

Ah, I see, the window closes if you have no changes to save, i.e. if the "Save Changes" button is not yet focusable.  It still seems pretty abrupt.

the-t-in-rtf:
> I think it's a good idea to show some warning to users that they're closing the panel without saving changes, and I'll address it in my next commit.

Great, as you can still lose unsaved changes either by clicking outside the dialogue or by hitting escape.

dmahajan980:
> Ah, I see, the window closes if you have no changes to save, i.e. if the "Save Changes" button is not yet focusable. It still seems pretty abrupt.

That's strange, as the panel does not close here even with the "Save Changes" button disabled. It jumps to the first configuration menu in both cases. I hope adding a warning to the user about the panel getting closed will solve the part.

PR #39, opened 2020-07-27
Added "Discard Changes" button to the configuration panel with the mechanism to store unsaved changes.

### LINKED ISSUE #38: Issues with the configuration panel when closed accidentally.
There are many cases where the user could accidentally close the panel without saving their changes. Adding a warning or saving the changes temporarily at least could solve the problem to a certain extent.

Upon digging further, I realized that it is not possible to keep the extension open or prevent it from closing, as their [FAQ says](https://developer.chrome.com/extensions/faq#faq-persist-popups). So the idea of a warning seems less likely, and the only possible way would be to store the changes separately. With this approach, we would have to create our mechanisms for clearing the data (if we want to store it temporarily), as the [Chrome Storage API](https://developer.chrome.com/extensions/storage) doesn't work like [sessionStorage](https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage).

To give a brief overview of the mechanism I'm thinking, it would store the changes side-by-side in Chrome's storage (separate from the stored configuration that the gamepad uses). So, if the user accidentally closes the panel, those changes would be reflected on the panel when opened again. In any case, the temporary changes would be deleted when the user saves the configuration using the "Save Changes" button.

While this might be the mechanism we later proceed with, I am open to other mechanisms, suggestions, and approaches that can be used.

dmahajan980:
Resolves #38 

After this pull request gets merged, the only thing for the configuration panel we are left with is adding a list for switching between configuration menus for various inputs.

### DIFF
--- a/src/js/configuration_panel/button-listeners.js
+++ b/src/js/configuration_panel/button-listeners.js
@@ -40,7 +40,7 @@ https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
         });
 
         // Toggle the "Save Changes" button.
-        that.toggleSaveButtonState();
+        that.toggleSaveAndDiscardButton();
     };
 
     /**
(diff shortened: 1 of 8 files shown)


src/js/configuration_panel/button-listeners.js
```diff
@@ -141,25 +201,39 @@ https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
...
      * @param {Object} that - The configurationPanel component.
      * @param {Object} saveChangesButton - The "Save Changes" button on the panel.
+     * @param {Object} discardButton - The "Discard Changes" button on the panel.
      *
      */
-    gamepad.configurationPanel.buttonListeners.toggleSaveButton = function (that, saveChangesButton) {
+    gamepad.configurationPanel.buttonListeners.toggleSaveAndDiscardButtons = function (that, saveChangesButton, discardButton) {
+        saveChangesButton = saveChangesButton[0];
+        discardButton = discardButton[0];
         chrome.storage.local.get(["gamepadConfiguration"], function (gamepadConfigurationWrapper) {
```
the-t-in-rtf, line 234: What I had in mind was to use the model listener infrastructure rather than checking to see if the state is "dirty" (unsaved changes) yourself.  I do like that you are able to tell if they undo their unsaved changes and get back to a "clean" state.  Let me look at the docs for a bit and think about how we might get both without having to check each time.

src/js/configuration_panel/button-listeners.js
```diff
@@ -176,50 +250,75 @@ https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
                         initialInputData = initialConfiguration[isAxes ? "axes" : "buttons"][inputIndex];
 
                     /**
-                     * Enable the "Save Changes" button if dropdown option is changed.
-                     * Otherwise, disable the button.
+                     * Enable the "Save Changes" and "Discard Changes" buttons if
+                     * dropdown option is changed.
                      */
                     var actionDropdown = configurationMenu.querySelector("select"),
```
the-t-in-rtf, line 256: Apologies for not catching this earlier.  Very soon we should refactor your work to use Fluid ViewComponents rather than direct selectors.  It would make it easier for someone to reuse your work with different markup.  I will write that up for later.

src/js/configuration_panel/button-listeners.js
the-t-in-rtf: OK, I talked to our architect at some length about this, he suggested keeping a copy of the last state saved to storage and then using [`fluid.model.diff`](https://github.com/fluid-project/infusion/blob/132aa848c1c55f88dac7e8ebd996c67bca8d482e/src/framework/core/js/DataBinding.js#L1425) to compare the current model to what's represented by the form.   This would tell you whether the new model has unsaved changes or not.

the-t-in-rtf: To make this work, you'd also need to make sure all form changes are relayed to the model.  We typically use [fluid-binder](https://github.com/fluid-project/fluid-binder) for this.

src/js/configuration_panel/button-listeners.js
```diff
@@ -141,25 +201,39 @@ https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
...
+     * @param {Object} discardButton - The "Discard Changes" button on the panel.
      *
      */
-    gamepad.configurationPanel.buttonListeners.toggleSaveButton = function (that, saveChangesButton) {
+    gamepad.configurationPanel.buttonListeners.toggleSaveAndDiscardButtons = function (that, saveChangesButton, discardButton) {
+        saveChangesButton = saveChangesButton[0];
+        discardButton = discardButton[0];
         chrome.storage.local.get(["gamepadConfiguration"], function (gamepadConfigurationWrapper) {
             // Get the list of all dropdowns in the configuration panel.
             var configurationMenus = document.querySelectorAll(".menu-item");
```
the-t-in-rtf, line 236: So, this is what I meant about needing `fluid-binder`, right now the DOM elements are the only place the state is stored until it gets saved.   It'd be better to have form changes relayed to a "working copy" model, IMO.

src/js/configuration_panel/button-listeners.js
```diff
@@ -176,50 +250,75 @@ https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
...
                         initialDropdownValue = initialInputData[isStoredData ? "currentAction" : "defaultAction"] || "null";
                     if (currentDropdownValue !== initialDropdownValue) {
                         saveChangesButton.removeAttribute("disabled");
+                        discardButton.removeAttribute("disabled");
+
+                        /**
+                         * Store the unsaved changes to avoid loss of configuration if
+                         * the panel is closed temporarily.
+                         */
+                        that.storeUnsavedChanges();
```
the-t-in-rtf, line 267: So the "unsaved changes" material would actually be a good candidate for a live relay, i.e. something that listens to model changes resulting from form updates and immediately saves them to Chrome Storage.  It would also need to listen to changes to Chrome Storage and relay those to the model.

I'm happy to defer on some of this, i.e. one pull to migrate to using view components, another to use fluid-binder, another to add model relay to/from Chrome Storage.

src/js/configuration_panel/button-listeners.js
the-t-in-rtf: Please add a TODO somewhere nearby referencing this ticket:

https://github.com/fluid-lab/gamepad-navigator/issues/40

src/js/configuration_panel/button-listeners.js
the-t-in-rtf: Please add a TODO somewhere nearby referencing this ticket:

https://github.com/fluid-lab/gamepad-navigator/issues/41

Issue #22, opened 2020-07-13
Add support for "long press" and standardise how actions make use of various inputs.


the-t-in-rtf:
Right now we have a bunch of approaches with at least some code duplicated between them to handle various types of input.  Once you've got the core functionality on board in some form (performing an action using at least one type of input), we should review and refactor and make each "style" of interaction into a clearer pattern and ensure that all functions can be used with:

1. Short presses that trigger an action on button release.
2. Long presses that trigger an action on button release after a configured time period.
2. Long presses that repeatedly trigger an action.

We'll need a kind of matrix for these, where you support each type of action for digital buttons, analogue buttons, and axes. This is probably one to handle later, for example, after you've got the configuration panel working.  It may be that we'll want to refactor to have single actions that support a direction and flags for the type of input, so that you don't have quite so many invokers or functions, but we can review closer to time and discuss.

the-t-in-rtf:
We should also think about how/whether we can support "chording" where a combination of buttons triggers an action rather than a single input.  For example, one key might act as a "reverse" modifier and change a button press from moving forward to moving backward.

duhrer:
I ran into related issues when working on #84.  Because we only track the button press, we can't do things like "hold to repeat", which makes navigation things like an on-screen keyboard pretty tedious.

Our current approach also makes it difficult to provide feedback when a button is held versus when it's released.   To address this (and make chording possible), we would need to:

1. Listen for button down, toggling the state of the button.
2. Periodically refire the event if the button is held.  The repeat rate should be configurable by the user, to avoid retriggering if someone needs more time to release a button.
3. Listen for button up (or infer button up on our state check).
4. For things like "click", we would need to trigger "mousedown", then "click" on button down, and "mouseup"  on release.

In rudimentary testing, I've confirmed that simulated mouse events other than click work:

```
var element = document.querySelector(".osk-key-Space");
elements.dispatchEvent(new MouseEvent("mouseup"));
```

This works in Firefox, Chrome, and Safari.

duhrer:
As far as I can see, this is a blocker for #77, and would really help with #84 and #85.

duhrer:
It would be good to look at the polling mechanism for the sticks and scrolling.

duhrer:
In the work leading up to the upcoming 1.0 release, I ended up refactoring a lot of the things I commented on above, notably the polling between repeats (see #108).

[...]

PR #42, opened 2020-07-28
Added feature to zoom in/out

### LINKED ISSUE #29: Add ability to zoom in/out.
It would be helpful to have the ability to zoom in and out (equivalent to ctrl+plus/minus on windows, command+plus/minus) on mac.

dmahajan980:
Resolves #29 

The page can now be zoomed in/out using gamepad's buttons and thumbsticks.

### DIFF
--- a/src/js/background_scripts/message-listener-utils.js
+++ b/src/js/background_scripts/message-listener-utils.js
@@ -139,4 +139,30 @@ https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
             }
         });
     };
+
+    /**
+     *
+     * Change the zoom value of the current browser tab.
+     *
+     * @param {String} zoomType - Determines if the page should be zoomed in or out.
+     *
+     */
+    gamepad.messageListenerUtils.setZoom = function (zoomType) {
+        chrome.tabs.query({ currentWindow: true, active: true }, function (currentTab) {
+            // Obtain the zoom value of the current tab.
+            chrome.tabs.getZoom(currentTab.id, function (currentZoomFactor) {
+                // Compute the new zoom value according to the zoom type.
+                var newZoomFactor = null;
+                if (zoomType === "zoomIn") {
+                    newZoomFactor = currentZoomFactor + 0.1;
+                }
+                else if (zoomType === "zoomOut") {
+                    newZoomFactor = currentZoomFactor - 0.1;
+                }
+
+                // Set the new zoom value.
+                chrome.tabs.setZoom(currentTab.id, newZoomFactor);
+            });
+        });
+    };
 })(fluid);
--- a/src/js/configuration_panel/configuration-panel.js
+++ b/src/js/configuration_panel/configuration-panel.js
@@ -85,14 +85,17 @@ https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
                 closeCurrentWindow: "Close current browser window",
                 openNewWindow: "Open a new browser window",
                 goToPreviousWindow: "Switch to the previous browser window",
-                goToNextWindow: "Switch to the next browser window"
+                goToNextWindow: "Switch to the next browser window",
+                zoomIn: "Zoom-in on the current tab",
+                zoomOut: "Zoom-out on the current tab"
             },
             axes: {
                 null: "None",
                 scrollHorizontally: "Scroll horizontally",
                 scrollVertically: "Scroll vertically",
                 thumbstickHistoryNavigation: "History navigation",
-                thumbstickTabbing: "Focus on the previous/next element"
+                thumbstickTabbing: "Focus on the previous/next element",
+                thumbstickZoom: "Zoom in or out on the current tab"
             }
         },
         // Describes the actions that use a particular configuration option.
@@ -113,7 +116,8 @@ https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
                 "scrollHorizontally",
                 "scrollVertically",
                 "thumbstickHistoryNavigation",
-                "thumbstickTabbing"
+                "thumbstickTabbing",
+                "thumbstickZoom"
             ]
         },
         invokers: {
--- a/src/js/configuration_panel/configuration-panel.js
+++ b/src/js/configuration_panel/configuration-panel.js
@@ -86,8 +86,8 @@ https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
                 openNewWindow: "Open a new browser window",
                 goToPreviousWindow: "Switch to the previous browser window",
                 goToNextWindow: "Switch to the next browser window",
-                zoomIn: "Zoom-in on the active webpage",
-                zoomOut: "Zoom-out on the active webpage"
+                zoomIn: "Zoom-in on the active web page",
+                zoomOut: "Zoom-out on the active web page"
             },
             axes: {
                 null: "None",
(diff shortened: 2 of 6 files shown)


the-t-in-rtf:
So, I tested this with an incidental button, the triggers, and both horizontal and vertical thumbstick axes.  This is exactly the kind of detailed testing we can easily do now that we have the configuration panel, it's a huge help with QA.

The incidental buttons are the most solid.  I wonder if we should avoid wrapping around, though.

The triggers are interesting, if you gradually vary the pressure, you can basically zoom in or out many times.  This is what I'd expect from the thumbsticks, so I guess it's fair to do the same for the triggers.   I would ask though that you put a TODO to discuss zooming when we look at #22, i.e. we might want to let people choose to use "incidental button" behaviour with the triggers.

The thumbsticks work well, although I found the natural ordering odd for the horizontal axes, i.e. left increased the zoom, right decreased it.  Probably the best way to handle that long term is to create an issue to add the ability to invert an axis.

src/js/background_scripts/message-listener-utils.js
```diff
@@ -139,4 +139,30 @@ https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
...
+     *
+     */
+    gamepad.messageListenerUtils.setZoom = function (zoomType) {
+        chrome.tabs.query({ currentWindow: true, active: true }, function (currentTab) {
+            // Obtain the zoom value of the current tab.
+            chrome.tabs.getZoom(currentTab.id, function (currentZoomFactor) {
+                // Compute the new zoom value according to the zoom type.
+                var newZoomFactor = null;
+                if (zoomType === "zoomIn") {
+                    newZoomFactor = currentZoomFactor + 0.1;
```
the-t-in-rtf, line 157: Was there some reason you didn't add a speed factor to the control panel and/or make this a configurable option?

src/js/content_scripts/input-mapper-background-utils.js
```diff
@@ -49,4 +49,30 @@ https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
...
+     * @param {Object} that - The inputMapper component.
+     * @param {Integer} value - The value of the gamepad input.
+     * @param {Boolean} invert - Whether the zooming should be in opposite order.
+     *
+     */
+    gamepad.inputMapperUtils.background.thumbstickZoom = function (that, value, invert) {
+        // Get the updated input value according to the configuration.
+        var inversionFactor = invert ? -1 : 1;
+        value = value * inversionFactor;
+        var zoomType = value > 0 ? "zoomOut" : "zoomIn",
```
the-t-in-rtf, line 67: I really suspect you're going to want some kind of cutoff for this, like 0.25 in either direction.  Many PS4 controllers especially have jitter in about that range, which would be really disconcerting in this case.  For now I'd say add another TODO to discuss when we work on #22.

src/js/background_scripts/message-listener-utils.js
```diff
@@ -139,4 +139,30 @@ https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
...
+                }
+                else if (zoomType === "zoomOut") {
+                    newZoomFactor = currentZoomFactor - 0.1;
+                }
+
+                // Set the new zoom value.
+                chrome.tabs.setZoom(currentTab.id, newZoomFactor);
```
the-t-in-rtf, line 164: I hadn't realised that modern browsers were doing exactly this, zooming the specific tab/window.

src/js/configuration_panel/configuration-panel.js
```diff
@@ -85,14 +85,17 @@ https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
...
+                zoomOut: "Zoom-out on the current tab"
             },
             axes: {
                 null: "None",
                 scrollHorizontally: "Scroll horizontally",
                 scrollVertically: "Scroll vertically",
                 thumbstickHistoryNavigation: "History navigation",
-                thumbstickTabbing: "Focus on the previous/next element"
+                thumbstickTabbing: "Focus on the previous/next element",
+                thumbstickZoom: "Zoom in or out on the current tab"
```
the-t-in-rtf, line 98: I'm wondering if this shouldn't be "tab/window", i.e. if there's only one tab.  I also think "active" is more helpful than "current".

src/js/background_scripts/message-listener-utils.js
dmahajan980: @the-t-in-rtf I didn't think of adding `speedFactor` while writing this because usually, users won't need anything to decide with what speed they want to zoom in (unlike the case we do have in scrolling and tab navigation where users might want to set up the navigation speed). I can add it if you think that'd be better.

src/js/content_scripts/input-mapper-background-utils.js
dmahajan980: I've added the cutoff check just below it. You can check that [here](https://github.com/fluid-lab/gamepad-navigator/pull/42/files/5022f4bccfa2fb5bd4cad25e2fc6210e166a57eb#diff-fee08b1acaa59bc549d236d3f4c9a80bR74). So the jittery stuff would be handled in any case.

src/js/configuration_panel/configuration-panel.js
dmahajan980: I think the "active webpage" would be better in that case. Should I replace the part of the description with it?

dmahajan980:
Hi @the-t-in-rtf, thank you for the feedback. I have some doubts regarding it, which I hope you'd clarify.

> I wonder if we should avoid wrapping around, though.

I'm not sure what you meant here. Could you please elaborate on it further?

> Probably the best way to handle that long term is to create an issue to add the ability to invert an axis.

Although the configuration panel already supports inverting the axes controls, I am not sure if you meant something else here. Please let me know if that's the case.

the-t-in-rtf:
> Although the configuration panel already supports inverting the axes controls, I am not sure if you meant something else here. Please let me know if that's the case.

Cheers, I missed it because I was looking at "Thumbstick Button" rather than "Thumbstick Horizontal Direction".

the-t-in-rtf:
> I wonder if we should avoid wrapping around, though.

So, if you keep zooming out beyond 0%, you end up at 100% again.  You can keep zooming in forever.

Anyway, the behaviour when zooming out matches what the keyboard does, so it's fine.

src/js/configuration_panel/configuration-panel.js
the-t-in-rtf: sounds good.

src/js/content_scripts/input-mapper-background-utils.js
the-t-in-rtf: Ah!  Thanks.

src/js/background_scripts/message-listener-utils.js
the-t-in-rtf: I think the increments are reasonable and fairly close to what the keyboard delivers.  Most of my concerns are with the extreme zooms generated by slowly depressing the triggers, and we can address that in #22.

src/js/configuration_panel/configuration-panel.js
```diff
@@ -86,16 +86,16 @@ https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
                 openNewWindow: "Open a new browser window",
                 goToPreviousWindow: "Switch to the previous browser window",
                 goToNextWindow: "Switch to the next browser window",
-                zoomIn: "Zoom-in on the current tab",
-                zoomOut: "Zoom-out on the current tab"
+                zoomIn: "Zoom-in on the active webpage",
```
the-t-in-rtf, line 89: Sorry to nitpick but "web page" is probably better.

dmahajan980: Whoops! My bad. I'll fix it now.

dmahajan980: @the-t-in-rtf it has been changed. Thanks for correcting me here.

PR #43, opened 2020-07-29
Added feature to maximize/restore windows

### LINKED ISSUE #28: Add ability to maximize / restore browser windows.
Browsers may open new windows/tabs that only cover part of the screen, we should provide a means of maximizing windows.  We should also provide a means of restoring the window size (typically the middle of the three window control icons).  Minimizing windows is out of scope for now.

dmahajan980:
Resolves #28 

The window size can be maximized and restored using both buttons and joysticks.

### DIFF
--- a/src/js/background_scripts/message-listener-utils.js
+++ b/src/js/background_scripts/message-listener-utils.js
@@ -223,7 +223,7 @@ https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
                             width: fluid.get(previousProperties, "width") || Math.round(3 * screen.width / 5),
                             height: fluid.get(previousProperties, "height") || Math.round(4 * screen.height / 5),
                             left: fluid.get(previousProperties, "left") || Math.round(screen.width / 15),
-                            top: fluid.get(previousProperties, "height") || Math.round(screen.height / 15)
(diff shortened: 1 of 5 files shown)


the-t-in-rtf:
I see the controls to do this, but can't seem to get it to work for any button or thumbstick axis, at least on the mac.

the-t-in-rtf:
The updated mac implementation works well.  As I discussed in chat, the only small issue is that if a window is in my second monitor and I maximise/restore it, it ends up on the primary monitor.

the-t-in-rtf:
I also notice that if I position two windows side by side to perform a comparison and then maximise/restore one, the previous position or size are not preserved.

dmahajan980:
@the-t-in-rtf as we discussed in our chat, I've updated the pull to restore the same dimensions before maximization but I couldn't test it for macOS. It may or may not work as expected. However, I've opened issue #45 for the same and I'll update it if the preservation of dimensions is working correctly.

the-t-in-rtf:
The position has definitely changed, but something is still off.  Restore consistently puts the window nearly offscreen (usually near the bottom of the screen).

the-t-in-rtf:
The position restore is really solid, good work.

PR #46, opened 2020-07-31
Added feature to reopen closed tabs/windows.

### LINKED ISSUE #24: Feature to reopen closed tabs using the gamepad
Currently, we can open and close tabs using the gamepad. There could be cases where the user accidentally closed a tab and might want to reopen it. So, a separate feature should be added, which would allow the users to open the closed tabs using the gamepad.

dmahajan980:
Resolves #24

### DIFF
--- a/src/js/configuration_panel/configuration-panel.js
+++ b/src/js/configuration_panel/configuration-panel.js
@@ -89,7 +89,8 @@ https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
                 zoomIn: "Zoom-in on the active web page",
                 zoomOut: "Zoom-out on the active web page",
                 maximizeWindow: "Maximize the current browser window",
-                restoreWindowSize: "Restore the size of current browser window"
+                restoreWindowSize: "Restore the size of current browser window",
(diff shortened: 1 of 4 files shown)


the-t-in-rtf:
I just finished the manual QA, this works as expected in every scenario I can imagine (except of course that it won't work if you close all windows, which is a known issue).  I will look through the code shortly.

src/js/configuration_panel/configuration-panel.js
```diff
@@ -89,7 +89,8 @@ https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
                 zoomIn: "Zoom-in on the active web page",
                 zoomOut: "Zoom-out on the active web page",
                 maximizeWindow: "Maximize the current browser window",
-                restoreWindowSize: "Restore the size of current browser window"
+                restoreWindowSize: "Restore the size of current browser window",
+                reopenTabOrWindow: "Re-open the closed tab or window"
```
the-t-in-rtf, line 93: I know that Chrome itself uses similar wording, but I'd prefer to add "last" here, as in "last closed".

PR #47, opened 2020-08-02
Fixed monitor switching issues with maximize/restore windows (on macOS X)

### LINKED ISSUE #45: Issues with maximize/restore windows in multiple monitor setup (macOS X)
The current approach to maximize/restore windows (in macOS X) requires setting up its dimensions and position offset. While this seems to work in single-monitor setups, the window (in the non-primary monitor) ends up on the primary monitor when maximized or restored using the gamepad. This might require inferring which monitor the browser is running on, and updating the position accordingly.

The other issue with the current implementation is that the window isn't restored to its original size (as before maximization). It might require preserving the dimensions of the window before maximization and re-using them when the window is restored.

dmahajan980:
Fixes #45

### DIFF
--- a/src/js/background_scripts/message-listener-utils.js
+++ b/src/js/background_scripts/message-listener-utils.js
@@ -208,7 +208,7 @@ https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
                         chrome.windows.update(windowPostUpdate.id, {
                             width: screen.width,
                             height: screen.height,
-                            left: 0,
+                            left: screen.availLeft,
                             top: 0
                         });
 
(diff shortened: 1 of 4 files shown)


the-t-in-rtf:
This fixes the "restore" action, i.e. the window is restored to the correct window.  However, a window on the second monitor is still maximised to the primary monitor.

src/js/background_scripts/message-listener-utils.js
```diff
@@ -222,7 +222,7 @@ https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
                         chrome.windows.update(windowPostUpdate.id, {
                             width: fluid.get(previousProperties, "width") || Math.round(3 * screen.width / 5),
                             height: fluid.get(previousProperties, "height") || Math.round(4 * screen.height / 5),
-                            left: fluid.get(previousProperties, "left") || Math.round(screen.width / 15),
+                            left: fluid.get(previousProperties, "left") || (screen.availLeft + Math.round(screen.width / 15)),
```
the-t-in-rtf, line 225: Seems like you just need add similar logic to the maximise action?

dmahajan980: Yes, I had replicated the same logic during maximization. But it seems that the background script is returning the equal value of `screen.availLeft` for all monitors (which should be different), and it produced the bug. So, I'm passing the value from content scripts now, and it seems to work.

dmahajan980:
@the-t-in-rtf I've updated the pull. Hopefully, it should work now. But please let me know if you're still facing any issues.

Issue #40, opened 2020-07-28
Migrate to using Fluid View Components instead of in-code jQuery selectors.


the-t-in-rtf:
Currently, much of the DOM inspection and manipulation relies on the use of hard-coded jQuery selectors.  This means that any refactor of the markup requires hunting through the code and ensuring that the selectors still match.  In other work we use [View Components](https://docs.fluidproject.org/infusion/development/tutorial-developerIntroduction/DeveloperIntroductionToInfusionFramework-ViewsAndViewComponents.html) and their selector mechanisms (`that.locate`).  This allows us to declaratively express the relationship between code and markup in a way that supports reuse and helps with refactoring.

As part of this, we should also discuss how/whether to break down the UI into smaller more manageable chunks, for example, one view component per option panel.

This work is a precursor to updating to relay component model data to/from form elements.

PR #48, opened 2020-08-02
Added list to switch between different inputs' configuration menu and polished the UI of the configuration panel

### LINKED ISSUE #30: Add configuration panel to remap inputs.
Currently the default input mapping is hard coded.  We need to provide a configuration panel to allow user to select which functions are mapped to which controls.

Ideally, we would like for this to be at least keyboard navigable, as we've discussed it doesn't necessarily need to be controllable using the joystick (although it would be nice if we could manage it).

dmahajan980:
Resolves #30

Along with the list view for the configuration panel, I've modified its UI, especially for the dark mode. If there's any preference for the previous interface or any specific part of the old UI, please let me know and I'll restore it.


the-t-in-rtf:
Based on our previous conversation, I thought you had something different in mind, for example a scrollable list of all buttons.    However, your design is a good improvement, as it greatly reduces the number of key strokes required to configure the later buttons in the list.   I'm happy to review the code and merge and see what feedback we get on the UI in the design critique.

src/js/configuration_panel/button-listeners.js
```diff
@@ -31,16 +31,16 @@ https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
      */
     gamepad.configurationPanel.buttonListeners.setAllToNone = function (that) {
         // Get the list of all dropdowns in the configuration panel.
-        var dropdownMenusArray = document.querySelectorAll("select");
```
the-t-in-rtf, line 34: Good idea.

src/js/configuration_panel/panel-events.js
```diff
@@ -59,7 +66,7 @@ https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
          */
         fluid.each(inputMenusArray, function (inputMenu) {
             if (fluid.isDOMNode(inputMenu)) {
-                that.changeConfigMenuOptions(inputMenu.querySelector("select"));
+                that.changeConfigMenuOptions(inputMenu.querySelector(".action-dropdown"));
```
the-t-in-rtf, line 69: I can see that you're kind of halfway using the viewComponent infrastructure.  When you get around to working on #40, you'll probably need to rework this to either render all the controls and show/hide them as needed, or otherwise avoid reusing a single set of controls for all inputs.  You'll also need to migrate to using `that.locate` rather than directly querying the DOM using jQuery.

If you could leave a TODO on the function that backs the `changeConfigMenuOptions` invoker, that'd be enough for now.

src/js/configuration_panel/panel-events.js
```diff
@@ -97,7 +104,7 @@ https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
      */
     gamepad.configurationPanel.handleEvents.changeConfigMenuOptions = function (that, dropdownMenu) {
         var selectedAction = $(dropdownMenu).val(),
-            dropdownClassName = dropdownMenu.classList[0],
+            dropdownClassName = dropdownMenu.classList[1],
```
the-t-in-rtf, line 107: If someone makes even small changes to your markup (such as adding their own class for style purposes), it will throw this off.  This can probably wait and be fixed as part of #40, but it needs to be fixed.  If you can leave a TODO here for now, that would be enough.

src/js/configuration_panel/panel-events.js
```diff
@@ -113,7 +120,7 @@ https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
             currentInputMenuItems[3].removeAttribute("disabled");
 
             // Reduce the width of the action dropdown.
-            dropdownMenu.style.width = "94%";
+            dropdownMenu.style.width = "96.5%";
```
the-t-in-rtf, line 123: Why do you have to do this here and not in the stylesheet?

src/js/configuration_panel/panel-events.js
dmahajan980: > you'll probably need to rework this to either render all the controls and show/hide them as needed, or otherwise avoid reusing a single set of controls for all inputs.

@the-t-in-rtf thanks for the feedback, but I'm slightly confused with this statement. The approach we are currently using does the same, i.e., render all the controls and display/hide them according to the action selected by the user. Also, I'm not sure what you mean by the second part of the statement. Could you please elaborate on this a bit?

src/js/configuration_panel/panel-events.js
dmahajan980: It was added here so that when the user switches to an action that shows up the "Speed Factor" input box, the width of the action dropdown is reduced accordingly.

src/js/configuration_panel/panel-events.js
the-t-in-rtf: I had mistakenly thought you were reusing a single element because of your strategy of determining which select it was using the classname.   I guess that instead you're using a general method for all selects and determining which you're working with using the classname.

In any case, my point is still that it would be better to make a sub-component for select elements in general, and to have each sub-component manage the mapping for one control and manage its visibility and settings.  The settings would be relayed into the parent component.

src/js/configuration_panel/panel-events.js
the-t-in-rtf: I see.  Could you make a toggle class called "expanded" and do everything else in the stylesheet?  That seems preferable to hard-specifying the width deep in code.

dmahajan980: @the-t-in-rtf that's a good idea. I'll make the changes.

PR #49, opened 2020-08-02
Added documentation for configMaps grade

dmahajan980:
This addresses the documentation for the `configMaps` grade used in the `configurationPanel` and the `inputMapper` components.

### DIFF
--- a/docs/components/configMaps.md
+++ b/docs/components/configMaps.md
@@ -12,7 +12,13 @@ https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
 
 # `gamepad.configMaps`
 
-// TODO: Add links referencing the "mix in" components' documentation.
+<!--
+TODO: Add links referencing the "mix in" components' documentation. 
+
+TODO: Add documentation for supported actions.
+Refer:
+https://github.com/fluid-lab/gamepad-navigator/issues/50
+-->
 
 This grade provides the configuration maps for the gamepad. It is used in the `gamepad.configurationPanel` and
 `gamepad.inputMapper.base` grades that provide the actual functionality according to the configuration.
--- a/docs/components/configMaps.md
+++ b/docs/components/configMaps.md
@@ -1,0 +1,146 @@
+<!--
+Copyright (c) 2020 The Gamepad Navigator Authors
+See the AUTHORS.md file at the top-level directory of this distribution and at
+https://github.com/fluid-lab/gamepad-navigator/raw/master/AUTHORS.md.
+
+Licensed under the BSD 3-Clause License. You may not use this file except in
+compliance with this License.
+
+You may obtain a copy of the BSD 3-Clause License at
+https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
+-->
+
+# `gamepad.configMaps`
+
+This grade provides the configuration maps for the gamepad in the form of component's model data.
+
+## Using this grade
+
+This can be used by passing it to the component's `gradeNames` option.
+
+``` javascript
+fluid.defaults("my.configMaps.grade", {
+    gradeNames: ["gamepad.configMaps"]
+};
+
+var configMapsInstance = my.configMaps.grade();
+```
+
+## Component Options
+
+This grade doesn't use or provide any configuration options. It only uses the model data to provide gamepad configuration
+maps to other components. However, the grade can be extended to provide custom configuration options.
+
+## Model Structure
+
+The grade stores the model data in the following format:
+
+``` javascript
+model: {
+    map: {
+        buttons: {
+            "0": {
+                defaultAction: "previousPageInHistory",
+                currentAction: null,
+                speedFactor: 1,
+                background: false
+            },
+            // The buttons below use same format as above but have different values for configuration options (keys).
+            "1": {...},
+            "2": {...},
+            "3": {...},
+            "4": {...},
+            "5": {...},
+            "6": {...},
+            "7": {...},
+            "8": {...},
+            "9": {...},
+            "10": {...},
+            "11": {...},
+            "12": {...},
+            "13": {...},
+            "14": {...},
+            "15": {...},
+            // The buttons below are reserved for special functionality so they cannot be configured.
+            "16": null,
+            "17": null
+        },
+        axes: {
+            "0": {
+                defaultAction: "scrollHorizontally",
+                currentAction: null,
+                speedFactor: 1,
+                invert: false
+            },
+            // The axes below use the same format as above but have different values for configuration options (keys).
+            "1": {...},
+            "2": {...},
+            "3": {...}
+        }
+    },
+    // Stores all the common configuration options for the gamepad actions.
+    commonConfiguration: {
+        homepageURL: "https://www.google.com/"
+    }
+}
+```
+
+The first 16 buttons and all the axes are configurable. The configuration options used in their model data are explained
+below:
+
+| Option | Supported Input | Description |
+| :---: | :---: | :--- |
+| `defaultAction` | `buttons` and `axes` | The default action triggered with the particular gamepad input if the `currentAction` is not provided. |
+| `currentAction` | `buttons` and `axes` | The action triggered with the particular gamepad input. It takes precedence over the `defaultAction` and is `null` by default. However, this can be specified when the component is created. |
+| `speedFactor` | `buttons` and `axes` | The speed with which the given action should be executed. It is only applicable if the action is continuous and hence supports only a few actions. |
+| `background` | `buttons` | Specifies whether the new tab or window should be opened in background and is applicable for those two actions only. |
+| `invert` | `axes` | Supported only for actions provided to `axes` and specifies whether the direction of the action should be inverted. |
+
+The option `homepageURL` inside `commonConfiguration` option specifies the URL of the webpage that should be loaded when
+the new tab or window is opened using the gamepad.
+
+Values of the above model options can be modified by using Infusion's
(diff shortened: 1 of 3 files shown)


docs/components/configMaps.md
```diff
@@ -0,0 +1,146 @@
...
+    }
+}
+```
+
+The first 16 buttons and all the axes are configurable. The configuration options used in their model data are explained
+below:
+
+| Option | Supported Input | Description |
+| :---: | :---: | :--- |
+| `defaultAction` | `buttons` and `axes` | The default action triggered with the particular gamepad input if the `currentAction` is not provided. |
```
the-t-in-rtf, line 93: You should add a section on supported actions.  This can be another page.

docs/components/configMaps.md
```diff
@@ -0,0 +1,146 @@
...
+| `currentAction` | `buttons` and `axes` | The action triggered with the particular gamepad input. It takes precedence over the `defaultAction` and is `null` by default. However, this can be specified when the component is created. |
+| `speedFactor` | `buttons` and `axes` | The speed with which the given action should be executed. It is only applicable if the action is continuous and hence supports only a few actions. |
+| `background` | `buttons` | Specifies whether the new tab or window should be opened in background and is applicable for those two actions only. |
```
the-t-in-rtf, line 96: "a new tab or window"

docs/components/configMaps.md
```diff
@@ -0,0 +1,146 @@
...
+Licensed under the BSD 3-Clause License. You may not use this file except in
+compliance with this License.
+
+You may obtain a copy of the BSD 3-Clause License at
+https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
+-->
+
+# `gamepad.configMaps`
+
+This grade provides the configuration maps for the gamepad in the form of component's model data.
```
the-t-in-rtf, line 15: We call this type of component a "mix in", i.e. you mix it in with one or more grades that provide actual functionality.  You don't have to mention that, but you should mention the grades it's meant to be mixed with.  If you want to handle that when writing those docs, leave a comment with a TODO, please.

docs/components/configMaps.md
```diff
@@ -0,0 +1,146 @@
...
+
+## Using this grade
+
+This can be used by passing it to the component's `gradeNames` option.
+
+``` javascript
```
the-t-in-rtf, line 21: You can fairly safely assume that people working in the Fluid ecosystem don't need this simple of an example.   IMO this would be a good place to demonstrate actually mixing it with a functional grade.

docs/components/configMaps.md
```diff
@@ -0,0 +1,146 @@
...
+## Component Options
+
+This grade doesn't use or provide any configuration options. It only uses the model data to provide gamepad configuration
+maps to other components. However, the grade can be extended to provide custom configuration options.
+
+## Model Structure
+
+The grade stores the model data in the following format:
+
+``` javascript
```
the-t-in-rtf, line 38: This should have failed the linting checks.  We typically use `snippet` as the code type for incomplete pieces of Javascript or JSON.

docs/components/configMaps.md
```diff
@@ -0,0 +1,146 @@
...
+fluid.defaults("my.configMaps.grade", {
+    gradeNames: ["gamepad.configMaps"]
+};
+
+var configMapsInstance = my.configMaps.grade();
+```
+
+## Component Options
+
+This grade doesn't use or provide any configuration options. It only uses the model data to provide gamepad configuration
```
the-t-in-rtf, line 31: It's more accurate to say that all configurable options for this component are also model variables.  You should also specifically document the home page URL, people extending or updating your work might want to change that.

docs/components/configMaps.md
```diff
@@ -0,0 +1,146 @@
...
+});
+
+// Modify maps using the changeApplier API.
+var configMapsInstanceOne = my.configMaps.grade();
+configMapsInstanceOne.applier.change("commonConfiguration.homepageURL", "https://www.github.com/");
+```
+
+## Invokers
+
+This grade doesn't provide any invoker methods.
```
the-t-in-rtf, line 146: "invokers" is fine.

docs/components/configMaps.md
the-t-in-rtf: Also, you should rework "in the form of component's model data".  One option is to just remove it, you cover the model data parts already in later paragraphs.

docs/components/configMaps.md
```diff
@@ -0,0 +1,146 @@
...
+            "7": {...},
+            "8": {...},
+            "9": {...},
+            "10": {...},
+            "11": {...},
+            "12": {...},
+            "13": {...},
+            "14": {...},
+            "15": {...},
+            // The buttons below are reserved for special functionality so they cannot be configured.
```
the-t-in-rtf, line 64: You don't have to write it out in the docs, but I'm curious what you have in mind.  I'm also curious if you have anything that has more than 16 buttons.

docs/components/configMaps.md
```diff
@@ -0,0 +1,146 @@
...
+
+| Option | Supported Input | Description |
+| :---: | :---: | :--- |
+| `defaultAction` | `buttons` and `axes` | The default action triggered with the particular gamepad input if the `currentAction` is not provided. |
+| `currentAction` | `buttons` and `axes` | The action triggered with the particular gamepad input. It takes precedence over the `defaultAction` and is `null` by default. However, this can be specified when the component is created. |
+| `speedFactor` | `buttons` and `axes` | The speed with which the given action should be executed. It is only applicable if the action is continuous and hence supports only a few actions. |
+| `background` | `buttons` | Specifies whether the new tab or window should be opened in background and is applicable for those two actions only. |
+| `invert` | `axes` | Supported only for actions provided to `axes` and specifies whether the direction of the action should be inverted. |
+
+The option `homepageURL` inside `commonConfiguration` option specifies the URL of the webpage that should be loaded when
```
the-t-in-rtf, line 99: I missed this on the first reading, it's important enough to highlight better.  I'd suggest making a table in the component options" section.  The first entry would be `model.commonConfiguration.homepageURL`.   The second would be `model.map`, summarising briefly what the map does and saying "see below for more details".

docs/components/configMaps.md
```diff
@@ -0,0 +1,146 @@
...
+the new tab or window is opened using the gamepad.
+
+Values of the above model options can be modified by using Infusion's
```
the-t-in-rtf, line 102: This can be simplified to just start with "The above model options".

docs/components/configMaps.md
the-t-in-rtf: See later comments on how I'd suggest handling that.

docs/components/configMaps.md
```diff
@@ -0,0 +1,146 @@
+<!--
```
the-t-in-rtf, line 1: I suspect that the linting checks were not running, and I was correct.  In your `Gruntfile.js`, you only check files in `docs`, and not anything in a subdirectory.  Please update that check to use `docs/**/*.js` and fix any linting errors.

docs/components/configMaps.md
dmahajan980: I thought of adding the controls to the README file. But I can add it here or as a separate page if you prefer.

docs/components/configMaps.md
dmahajan980: Thanks, @the-t-in-rtf for pointing this out. I have updated it.

docs/components/configMaps.md
dmahajan980: The `Button 16` is the "badge icon," which we were planning to use for opening the configuration panel if the Chrome extensions allow us to do so (as we discussed in one of our email conversations). And the `Button 17` is the touchpad which acts as a mouse in some operating systems (at least in Ubuntu), so I isolated these buttons. But should I remove this part (comment and model variables) from the docs or keep it as it currently is?

docs/components/configMaps.md
dmahajan980: Thanks, I've updated the `Gruntfile` accordingly.

docs/components/configMaps.md
the-t-in-rtf: If you're going to document the actions in a separate file/ticket, that's fine, then just file an issue and leave a TODO here.

docs/components/configMaps.md
the-t-in-rtf: Thanks, somehow I had the idea that 15 was the badge, and I haven't hooked up my PS4 Dualshock in a while, so I forgot about the touchpad.  I think it's fine to hold on to it for special features.

docs/components/configMaps.md
```diff
@@ -12,30 +12,38 @@ https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
 
 # `gamepad.configMaps`
 
-This grade provides the configuration maps for the gamepad in the form of component's model data.
+// TODO: Add links referencing the "mix in" components' documentation.
```
the-t-in-rtf, line 15: You'll need to use HTML comment syntax to avoid having the comment show up in the rendered output.

docs/components/navigator.md
```diff
@@ -28,7 +28,7 @@ var navigatorInstanceOne = gamepad.navigator();
 // Otherwise create a custom component using the navigator grade.
 fluid.defaults("my.navigator.grade", {
     gradeNames: ["gamepad.navigator"]
-};
+});
```
the-t-in-rtf, line 31: I totally missed that.  This is the kind of thing that linting checks are really good at.

PR #51, opened 2020-08-05
Added documentation for inputMapper.base component

dmahajan980:
While writing this document, I discovered a few things which should be fixed. I've added TODOs for now, which can be addressed later.

### DIFF
--- a/docs/components/inputMapper.base.md
+++ b/docs/components/inputMapper.base.md
@@ -22,14 +22,10 @@ for **inter-web page navigation** features)
 
 ## Using this grade
 
-For the actions to work, you need to use the `gamepad.inputMapperUtils.content` namespace unless the `gamepad` namespace
-is used. Otherwise, the component will throw errors. The configuration options can be changed by either extending the
-component or by passing your own options. If no options are passed, the defaults are used.
+The configuration options can be changed by either extending the component or by passing your own options. If no
+options are passed, the defaults are used.
 
 ``` javascript
-fluid.registerNamespace("gamepad.inputMapper.base");
-fluid.registerNamespace("gamepad.inputMapperUtils.content");
-
 // Either create an instance of the inputMapper.base component.
 var inputMapperBaseInstanceOne = gamepad.inputMapper.base();
 
@@ -52,9 +48,6 @@ The following component configuration options are supported:
 These options can be provided to the component in the following ways:
 
 ``` javascript
-fluid.registerNamespace("gamepad.inputMapper.base");
-fluid.registerNamespace("gamepad.inputMapperUtils.content");
-
 // Either pass options inside the object as an argument while creating an instance of inputMapper.base component.
 var inputMapperBaseInstanceOne = gamepad.inputMapper.base({
     cutoffValue: 0.15,
@@ -76,9 +69,6 @@ to the [Component Options](configMaps.md#component-options) section in the [`con
 documentation.
 
 ``` javascript
-fluid.registerNamespace("gamepad.inputMapper.base");
-fluid.registerNamespace("gamepad.inputMapperUtils.content");
-
 var inputMapperBaseInstance = gamepad.inputMapper.base({
     cutoffValue: 0.35,
     scrollInputMultiplier: 60,
--- a/docs/components/inputMapper.base.md
+++ b/docs/components/inputMapper.base.md
@@ -16,9 +16,9 @@ https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
 
 This component transforms the gamepad inputs into actions. It is derived from the combination of two grades - the
 [`navigator`](navigator.md) grade, which reads and stores the gamepad inputs in the form of model data, and the
-[`configMaps`](configMaps.md) grade provides a configuration map for the gamepad inputs. The navigation features
-available with this component are limited to the **intra-web page navigation** features. (Use the `inputMapper`
-component for **inter-web page navigation** features)
+[`configMaps`](configMaps.md) grade, which provides a configuration map for the gamepad inputs. The navigation features
+available with this component are limited to **intra-web page navigation** features. (Use the `inputMapper` component
+for **inter-web page navigation** features)
 
 ## Using this grade
 
@@ -70,9 +70,10 @@ fluid.defaults("my.inputMapperBase.grade", {
 var inputMapperBaseInstanceTwo = my.inputMapperBase.grade();
 ```
 
-You can also pass custom configuration map (in the form of model variables) to specify the action and other associated
+You can also pass a custom configuration map (in the form of model variables) to specify the action and other associated
 parameters for the gamepad inputs as an argument. For information regarding the configuration map model variables, refer
-to the [Component Options](configMaps.md#component-options) section in the [`configMaps`](configMaps.md) grade document.
+to the [Component Options](configMaps.md#component-options) section in the [`configMaps`](configMaps.md) grade
+documentation.
 
 ``` javascript
 fluid.registerNamespace("gamepad.inputMapper.base");
@@ -96,15 +97,14 @@ var inputMapperBaseInstance = gamepad.inputMapper.base({
 
 ## Non-navigation Invokers
 
-### `{inputMapper.base}.produceNavigation(that, change)`
+### `{inputMapper.base}.produceNavigation(change)`
 
-- `that {Object}` The `inputMapper.base` component.
 - `change {Object}` Recipt for the change in input values.
 - Returns: Nothing.
 
 Listens for the changes in gamepad's input values stored as `inputMapper.base` component's model data and calls the
 **navigation-producing invokers** according to the configuration map. The navigation-producing invoker is triggered only
-if the component provides it. Each of these invokers is passed with the following arguments:
+if the component provides it. Each of these invokers accepts the following arguments:
 
 - `inputValue`: Current value of the gamepad input.
 - `speedFactor`: Times by which the speed of given (continuous) action should be increased.
@@ -113,42 +113,42 @@ if the component provides it. Each of these invokers is passed with the followin
 - `oldInputValue`: Previous value of the gamepad input before its state is changed.
 - `homepageURL`: URL that a new tab or window should load when opened.
 
-The invokers use only those arguments required to perform the navigation it's meant for. Apart from the `inputValue` and
+The invokers use only those arguments required to perform their specific action. Apart from the `inputValue` and
 `oldInputValue` arguments, the remaining arguments derive their value from the configuration map.
 
-### `{inputMapper.base}.clearIntervalRecords(records)`
+### `{inputMapper.base}.clearIntervalRecords()`
 
-- `records {Object}` The `intervalRecords` member containing a list of interval loops.
 - Returns: Nothing.
 
-Clears all the running interval loops when the gamepad is disconnected or when the instance of the `inputMapper.base`
+Clears all running interval loops when the gamepad is disconnected or when the instance of the `inputMapper.base`
 component is destroyed.
 
 ### `{inputMapper.base}.tabindexSortFilter(elementOne, elementTwo)`
 
-- `elementOne {Object}` The DOM element to be compared and sorted.
-- `elementTwo {Object}` The DOM element to be compared and sorted.
-- Returns: `Integer` An integer determining the order of two elements.
+- `elementOne {Object}` A DOM element to be compared and sorted.
+- `elementTwo {Object}` Another DOM element to be compared and sorted.
+- Returns: `Integer` An integer determining the order of two elements. Refer to the
+  [sort method](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort#Description)
+  documentation for possible values.
 
 Sorts all the tabbable elements in the order of their `tabindex` value. If none of the elements have a `tabindex`
-attribute then the elements are placed in the given order.
+attribute then the elements are placed in the order the elements appear in the DOM.
 
-### `{inputMapper.base}.trackDOM(that)`
+### `{inputMapper.base}.trackDOM()`
 
-- `that {Object}` The `inputMapper.base` component.
 - Returns: Nothing.
 
 Listens for changes to the DOM using `MutationObserver`. When the DOM has loaded or is changed, it creates a list of
 tabbable elements, sorts them using
 [`{inputMapper.base}.tabindexSortFilter`](#inputmapperbasetabindexsortfilterelementone-elementtwo), and then stores them
-the `tabbableElements` member variable for use during tab navigation.
+in the `tabbableElements` member variable for use during tab navigation.
 
 ### `{inputMapper.base}.stopTrackingDOM()`
 
 - Returns: Nothing.
 
-Stops the `MutationObserver` called by the [`{inputMapper.base}.trackDOM`](#inputmapperbasetrackdomthat) when the
-instance of the `inputMapper.base` component is destroyed. It avoids listening changes to the DOM any further.
+Stops the `MutationObserver` started by the [`{inputMapper.base}.trackDOM`](#inputmapperbasetrackdom) from listening to
+changes to the DOM. Called by the `inputMapper.base` component when it is destroyed.
 
 ## Navigation Invokers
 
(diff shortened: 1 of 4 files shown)


docs/components/inputMapper.base.md
```diff
@@ -0,0 +1,300 @@
...
+https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
+-->
+
+# `gamepad.inputMapper.base`
+
+<!-- TODO: Add links to the inputMapper component's documentation -->
+
+This component transforms the gamepad inputs into actions. It is derived from the combination of two grades - the
+[`navigator`](navigator.md) grade, which reads and stores the gamepad inputs in the form of model data, and the
+[`configMaps`](configMaps.md) grade provides a configuration map for the gamepad inputs. The navigation features
```
the-t-in-rtf, line 19: This should probably match the first part of the sentence, i.e. "grade, which provides".

docs/components/inputMapper.base.md
```diff
@@ -0,0 +1,300 @@
...
+available with this component are limited to the **intra-web page navigation** features. (Use the `inputMapper`
```
the-t-in-rtf, line 20: you can leave out the "the" before "intra-web".

docs/components/inputMapper.base.md
```diff
@@ -0,0 +1,300 @@
...
+component for **inter-web page navigation** features)
+
+## Using this grade
+
+For the actions to work, you need to use the `gamepad.inputMapperUtils.content` namespace unless the `gamepad` namespace
```
the-t-in-rtf, line 25: I'm not sure I get what you're going for here.  Can you describe in a comment what you were trying to say?

docs/components/inputMapper.base.md
```diff
@@ -0,0 +1,300 @@
...
+// Otherwise pass it as default options in a custom component using the inputMapper.base grade.
+fluid.defaults("my.inputMapperBase.grade", {
+    gradeNames: ["gamepad.inputMapper.base"],
+    cutoffValue: 0.10,
+    scrollInputMultiplier: 25
+});
+var inputMapperBaseInstanceTwo = my.inputMapperBase.grade();
+```
+
+You can also pass custom configuration map (in the form of model variables) to specify the action and other associated
```
the-t-in-rtf, line 73: "a custom configuration map".

docs/components/inputMapper.base.md
```diff
@@ -0,0 +1,300 @@
...
+parameters for the gamepad inputs as an argument. For information regarding the configuration map model variables, refer
+to the [Component Options](configMaps.md#component-options) section in the [`configMaps`](configMaps.md) grade document.
```
the-t-in-rtf, line 75: "document" is technically correct, but I usually prefer "documentation" in cases like this.

docs/components/inputMapper.base.md
```diff
@@ -0,0 +1,300 @@
...
+
+### `{inputMapper.base}.produceNavigation(that, change)`
+
+- `that {Object}` The `inputMapper.base` component.
+- `change {Object}` Recipt for the change in input values.
+- Returns: Nothing.
+
+Listens for the changes in gamepad's input values stored as `inputMapper.base` component's model data and calls the
+**navigation-producing invokers** according to the configuration map. The navigation-producing invoker is triggered only
+if the component provides it. Each of these invokers is passed with the following arguments:
```
the-t-in-rtf, line 107: probably "is passed with" should be simplified to "accepts".

docs/components/inputMapper.base.md
```diff
@@ -0,0 +1,300 @@
...
+
+- `inputValue`: Current value of the gamepad input.
+- `speedFactor`: Times by which the speed of given (continuous) action should be increased.
+- `invert`: Whether direction of the given action should be in the opposite order (for thumbsticks).
+- `background`: Whether a new tab or window should open in background.
+- `oldInputValue`: Previous value of the gamepad input before its state is changed.
+- `homepageURL`: URL that a new tab or window should load when opened.
+
+The invokers use only those arguments required to perform the navigation it's meant for. Apart from the `inputValue` and
```
the-t-in-rtf, line 116: I see what you mean, but "the navigation it's meant for" is a bit awkward.  I'd suggest "their specific action".

docs/components/inputMapper.base.md
```diff
@@ -0,0 +1,300 @@
...
+`oldInputValue` arguments, the remaining arguments derive their value from the configuration map.
+
+### `{inputMapper.base}.clearIntervalRecords(records)`
+
+- `records {Object}` The `intervalRecords` member containing a list of interval loops.
+- Returns: Nothing.
+
+Clears all the running interval loops when the gamepad is disconnected or when the instance of the `inputMapper.base`
```
the-t-in-rtf, line 124: "all running" is better here.

docs/components/inputMapper.base.md
```diff
@@ -0,0 +1,300 @@
...
+component is destroyed.
+
+### `{inputMapper.base}.tabindexSortFilter(elementOne, elementTwo)`
+
+- `elementOne {Object}` The DOM element to be compared and sorted.
+- `elementTwo {Object}` The DOM element to be compared and sorted.
+- Returns: `Integer` An integer determining the order of two elements.
```
the-t-in-rtf, line 131: You should describe the results, i.e. -1 means `elementOne` should appear first, et cetera.  You might also link to documentation for array sort, which I assume is the convention you're following.

docs/components/inputMapper.base.md
```diff
@@ -0,0 +1,300 @@
...
+- `records {Object}` The `intervalRecords` member containing a list of interval loops.
+- Returns: Nothing.
+
+Clears all the running interval loops when the gamepad is disconnected or when the instance of the `inputMapper.base`
+component is destroyed.
+
+### `{inputMapper.base}.tabindexSortFilter(elementOne, elementTwo)`
+
+- `elementOne {Object}` The DOM element to be compared and sorted.
+- `elementTwo {Object}` The DOM element to be compared and sorted.
```
the-t-in-rtf, line 130: These two should be distinguished somehow, for example "An element to be compared and sorted", and "Another element...".

docs/components/inputMapper.base.md
```diff
@@ -0,0 +1,300 @@
...
+- Returns: `Integer` An integer determining the order of two elements.
+
+Sorts all the tabbable elements in the order of their `tabindex` value. If none of the elements have a `tabindex`
+attribute then the elements are placed in the given order.
```
the-t-in-rtf, line 134: I know what you mean, but you should spell out "given order" better, as it's not obvious to people who haven't work with tabindex values and keyboard navigation.  I think "in the order the elements appear in the DOM" is good.

docs/components/inputMapper.base.md
```diff
@@ -0,0 +1,300 @@
...
+
+### `{inputMapper.base}.trackDOM(that)`
+
+- `that {Object}` The `inputMapper.base` component.
+- Returns: Nothing.
+
+Listens for changes to the DOM using `MutationObserver`. When the DOM has loaded or is changed, it creates a list of
+tabbable elements, sorts them using
+[`{inputMapper.base}.tabindexSortFilter`](#inputmapperbasetabindexsortfilterelementone-elementtwo), and then stores them
+the `tabbableElements` member variable for use during tab navigation.
```
the-t-in-rtf, line 144: Seems like you're missing "in" before "the `tabbableElements`".

docs/components/inputMapper.base.md
```diff
@@ -0,0 +1,300 @@
...
+
+### `{inputMapper.base}.stopTrackingDOM()`
+
+- Returns: Nothing.
+
+Stops the `MutationObserver` called by the [`{inputMapper.base}.trackDOM`](#inputmapperbasetrackdomthat) when the
+instance of the `inputMapper.base` component is destroyed. It avoids listening changes to the DOM any further.
```
the-t-in-rtf, line 151: This paragraph needs help, you have to search for a bit to figure out what you mean by "it" (the action) in the last sentence.  I'd suggest "Stops the MO started by (invoker link) from listening to changes to the DOM. Called by the (gradename) component when it is destroyed."

docs/components/inputMapper.base.md
```diff
@@ -0,0 +1,300 @@
...
+
+## Navigation Invokers
+
+### `{inputMapper.base}.click(value)`
+
+- `value {Number}` Current value of the gamepad input.
+- Returns: Nothing.
+
+Performs **click** on the focused element when the gamepad `button` or `trigger` is pressed. However, the `select`
```
the-t-in-rtf, line 160: I think "performs a click" would probably be better.

the-t-in-rtf, line 160: I think "the `select`" is probably not quite right, as you want to make it clear that the behaviour applies to all select form elements.  I'd suggest something like "a click on a select form element is simulated by changing its size".

docs/components/inputMapper.base.md
```diff
@@ -0,0 +1,300 @@
...
+dropdown is toggled by changing its `size` attribute when clicked using the gamepad.
```
the-t-in-rtf, line 161: Given that everything the extension does uses the gamepad, you can leave that phrase out.

docs/components/inputMapper.base.md
```diff
@@ -0,0 +1,300 @@
...
+### `{inputMapper.base}.previousPageInHistory(that, value)`
+
+- `that {Object}` The `inputMapper.base` component.
+- `value {Number}` Current value of the gamepad input.
+- Returns: Nothing.
+
+Navigates to the previous page in history when the gamepad `button` or `trigger` is pressed. It uses the
+[`History API`](https://developer.mozilla.org/en-US/docs/Web/API/History_API) for the core functionality. The `focused
+element` of current web page is saved using [`chrome.storage`](https://developer.chrome.com/extensions/storage) before
+the operation is performed. It is used for restoring focus on the same element when the user navigates back to the same
```
the-t-in-rtf, line 172: I would spell out "it" here, i.e. "the saved value is used to restore focus" or something like that.

docs/components/inputMapper.base.md
```diff
@@ -0,0 +1,300 @@
...
+page.
+
+### `{inputMapper.base}.nextPageInHistory(that, value)`
+
+- `that {Object}` The `inputMapper.base` component.
+- `value {Number}` Current value of the gamepad input.
+- Returns: Nothing.
+
+Navigates to the next page in history when the gamepad `button` or `trigger` is pressed. It uses the
```
the-t-in-rtf, line 181: This is one case where I think the backticks aren't needed, i.e. button and trigger are meaningful words, but not grade names, methods, or variables.  If your arguments had a `button` or `trigger` variable, you would only use the backticks if you were referring to the variable and not the concept of a trigger/button.

docs/components/inputMapper.base.md
```diff
@@ -0,0 +1,300 @@
...
+[`History API`](https://developer.mozilla.org/en-US/docs/Web/API/History_API) for the core functionality. The `focused
+element` of current web page is saved using [`chrome.storage`](https://developer.chrome.com/extensions/storage) before
+the operation is performed. It is used for restoring focus on the same element if the user navigates back to the same
```
the-t-in-rtf, line 184: Same comment here about improving "it".

docs/components/inputMapper.base.md
```diff
@@ -0,0 +1,300 @@
...
+page.
+
+### `{inputMapper.base}.thumbstickHistoryNavigation(that, value, invert)`
+
+- `that {Object}` The `inputMapper.base` component.
+- `value {Number}` Current value of the gamepad input.
+- `invert {Boolean}` Whether the thumbstick history navigation should be in opposite order.
```
the-t-in-rtf, line 191: All of these methods that support `invert` should probably describe what the default and inverted order actually mean.  This could be handled in the description of the function and you could just put "(see below)" in the list of arguments.

docs/components/inputMapper.base.md
```diff
@@ -0,0 +1,300 @@
...
+
+Calls the [`{inputMapper.base}.previousPageInHistory`](#inputmapperbasepreviouspageinhistorythat-value) and
+[`{inputMapper.base}.nextPageInHistory`](#inputmapperbasenextpageinhistorythat-value) invokers according to the
+direction the `thumbstick` is pressed. The direction can be inverted by passing `true` as the `invert` argument.
+
+### `{inputMapper.base}.reverseTab(that, value, direction)`
+
+- `that {Object}` The `inputMapper.base` component.
+- `value {Number}` Current value of the gamepad input.
+- `direction {String}` The direction for backward tab navigation, i.e. `reverseTab`.
```
the-t-in-rtf, line 202: This seems wrong, i.e. the `direction` should be hard-coded in the invoker's definition and shouldn't be accepted as an argument.

docs/components/inputMapper.base.md
```diff
@@ -0,0 +1,300 @@
...
+
+Changes the focus from the currently focused element to the previous tabbable element using gamepad `buttons` and
+`triggers`. The previous tabbable element is the element before the currently focused element in the array stored as
+`tabbableElements` member variable.
+
+### `{inputMapper.base}.forwardTab(that, value, direction)`
+
+- `that {Object}` The `inputMapper.base` component.
+- `value {Number}` Current value of the gamepad input.
+- `direction {String}` The direction for forward tab navigation, i.e. `forwardTab`.
```
the-t-in-rtf, line 213: Same comment here about the `direction`.

docs/components/inputMapper.base.md
```diff
@@ -0,0 +1,300 @@
...
+Changes the focus from the currently focused element to the next tabbable element using gamepad `buttons` and
+`triggers`. The next tabbable element is the element next to the currently focused element in the array stored as
+`tabbableElements` member variable.
+
+### `{inputMapper.base}.thumbstickTabbing(that, value, speedFactor, invert)`
+
+- `that {Object}` The `inputMapper.base` component.
+- `value {Number}` Current value of the gamepad input.
+- `speedFactor {Number}` Times by which the speed of tab navigation should be increased.
+- `invert {Boolean}` Whether the thumbstick tab navigation should be in opposite order.
```
the-t-in-rtf, line 225: Same comment here about needing to know what the default order is before you can understand the inverse.  You can use a simple single-sentence example, "for example, left on the horizontal axis of a thumbstick...".

docs/components/inputMapper.base.md
```diff
@@ -0,0 +1,300 @@
...
+
+Changes the focus from the currently focused element to the next tabbable element using gamepad `buttons` and
+`triggers`. The next tabbable element is the element next to the currently focused element in the array stored as
+`tabbableElements` member variable.
+
+### `{inputMapper.base}.thumbstickTabbing(that, value, speedFactor, invert)`
+
+- `that {Object}` The `inputMapper.base` component.
+- `value {Number}` Current value of the gamepad input.
+- `speedFactor {Number}` Times by which the speed of tab navigation should be increased.
```
the-t-in-rtf, line 224: This sentence can be improved, I suspect you should describe this in terms of frequency.  If there are units (for example, milliseconds), it's good to mention them.

docs/components/inputMapper.base.md
```diff
@@ -0,0 +1,300 @@
...
+- `invert {Boolean}` Whether the thumbstick tab navigation should be in opposite order.
+- Returns: Nothing.
+
+Changes the focus from the currently focused element to the previous or next tabbable element using gamepad
+`thumbsticks`. It calls the [`{inputMapper.base}.reverseTab`](#inputmapperbasereversetabthat-value-direction) and
```
the-t-in-rtf, line 229: You can avoid using "It" here and just say "Calls".

docs/components/inputMapper.base.md
```diff
@@ -0,0 +1,300 @@
...
+[`{inputMapper.base}.forwardTab`](#inputmapperbaseforwardtabthat-value-direction) invokers according to the direction
+the thumbstick is pressed and works only in **continuous** mode. The direction can be inverted by passing `true` as
```
the-t-in-rtf, line 231: You should probably make a separate sentence out of "works only in...".

docs/components/inputMapper.base.md
```diff
@@ -0,0 +1,300 @@
...
+the `invert` argument, and the speed of tab navigation can be changed by changing the value of `speedFactor` argument.
```
the-t-in-rtf, line 232: I see a lot of overlap between your descriptions and the variables/arguments.  I would suggest using the description to cover information about the function's behaviour in general, and specifics that are too long for the argument signature.   In any case you shouldn't be spelling out the same thing in two places so close to each other.

docs/components/inputMapper.base.md
```diff
@@ -0,0 +1,300 @@
...
+
+### `{inputMapper.base}.scrollLeft(that, value, speedFactor)`
+
+- `that {Object}` The `inputMapper.base` component.
+- `value {Number}` Current value of the gamepad input.
+- `speedFactor {Number}` Times by which the speed of left scrolling should be increased.
+- Returns: Nothing.
+
+Scrolls the web page in left direction using gamepad `buttons` and `triggers`. The speed of scrolling can be changed by
```
the-t-in-rtf, line 241: "in left direction" can just be "left".  The second sentence is a perfect example of information already covered in the variable list, and can be removed.

docs/components/inputMapper.base.md
```diff
@@ -0,0 +1,300 @@
...
+modifying the value of `speedFactor` argument.
+
+### `{inputMapper.base}.scrollRight(that, value, speedFactor)`
+
+- `that {Object}` The `inputMapper.base` component.
+- `value {Number}` Current value of the gamepad input.
+- `speedFactor {Number}` Times by which the speed of right scrolling should be increased.
+- Returns: Nothing.
+
+Scrolls the web page in right direction using gamepad `buttons` and `triggers`. The speed of scrolling can be changed by
```
the-t-in-rtf, line 251: "in right direction" can just be "right".

docs/components/inputMapper.base.md
```diff
@@ -0,0 +1,300 @@
...
+modifying the value of `speedFactor` argument.
+
+### `{inputMapper.base}.scrollUp(that, value, speedFactor)`
+
+- `that {Object}` The `inputMapper.base` component.
+- `value {Number}` Current value of the gamepad input.
+- `speedFactor {Number}` Times by which the speed of upward scrolling should be increased.
+- Returns: Nothing.
+
+Scrolls the web page in upward direction using gamepad `buttons` and `triggers`. The speed of scrolling can be changed
```
the-t-in-rtf, line 261: All of these "in X direction" can generally remove "in" and "direction".   Also, fix `buttons` and `triggers` throughout.

docs/components/inputMapper.base.md
```diff
@@ -0,0 +1,300 @@
...
+
+### `{inputMapper.base}.scrollHorizontally(that, value, speedFactor, invert)`
+
+- `that {Object}` The `inputMapper.base` component.
+- `value {Number}` Current value of the gamepad input.
+- `speedFactor {Number}` Times by which the speed of horizontal scrolling should be increased.
+- `invert {Boolean}` Whether the horizontal scrolling should be in opposite order.
+- Returns: Nothing.
+
+Scrolls the web page in left and right direction using gamepad `thumbsticks`. It calls the
```
the-t-in-rtf, line 282: `thumbsticks` should also just be thumbsticks (no backticks).

src/js/content_scripts/input-mapper-background-utils.js
```diff
@@ -20,6 +20,11 @@ https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
     // TODO: Add continuous / long-press browser tab navigation, if needed.
     // TODO: Add browser tab navigation for thumbsticks.
 
+    /**
+     * TODO: Save the DOM element focused before switching tabs so that it can be
```
the-t-in-rtf, line 24: probably just "focus" here makes more sense.

docs/components/inputMapper.base.md
dmahajan980: The invoker functions for navigation use the namespace `gamepad.inputMapperUtils.content`. So, I added that they need to use this namespace for navigation in cases where they don't register the `gamepad` namespace.

docs/components/inputMapper.base.md
dmahajan980: @the-t-in-rtf it's indeed hard-coded. But removing it raises another question about removing `that` since it's specified/passed in the invoker's `args` option and not when the invoker is called.

dmahajan980: @the-t-in-rtf thanks for pointing this out. I think I made a mistake with the arguments list here and as well as in the `navigator` documentation as it seems we should only specify those arguments which the user can provide and not the ones used in the invoker's definition.

docs/components/inputMapper.base.md
dmahajan980: The `speedFactor` isn't frequency. It is currently used as a multiplication factor for the frequency and has no units.

docs/components/inputMapper.base.md
the-t-in-rtf: Yes, I didn't even notice that `that` was an argument, but you've got the idea now.

docs/components/inputMapper.base.md
the-t-in-rtf: I guess I assumed that including your work was enough to define that as far as `fluid.invokeGlobal` needs (that backs the `funcName` stuff IIRC).  But I guess you have tried this with a different namespace and had problems?

docs/components/inputMapper.base.md
the-t-in-rtf: If I understand you correctly, you're multiplying the frequency (polling rate) by this?  If so, you should describe what is being scaled.  So, you might say "A scaling factor for the polling frequency, which affects the scrolling speed.  The default is to poll every X ms, setting this to less than 1 will slow down the scrolling, setting it to more than 1 will speed up the scrolling."  Something like that.

docs/components/inputMapper.base.md
the-t-in-rtf: Looks better, but I forgot to mention that click probably shouldn't be bold.

docs/components/inputMapper.base.md
the-t-in-rtf: "focused element" probably shouldn't be bold.

the-t-in-rtf: Sorry, I just noticed the backticks around "chrome.storage" and "History API".  We should reserve backticks for variables, gradeNames, invokers, i.e. things referenced in actual code.

docs/components/inputMapper.base.md
```diff
@@ -0,0 +1,282 @@
...
+Scrolls the web page downward using gamepad buttons and triggers.
+
+### `{inputMapper.base}.scrollHorizontally(value, speedFactor, invert)`
+
+- `value {Number}` Current value of the gamepad input.
+- `speedFactor {Number}` Times by which the speed of horizontal scrolling should be increased.
+- `invert {Boolean}` Whether the horizontal scrolling should be in opposite order (see below).
+- Returns: Nothing.
+
+Scrolls the web page in left and right direction using gamepad thumbsticks. Calls the
```
the-t-in-rtf, line 265: I missed this variation before, this should just be "scrolls the page left and right using the gamepad thumbsticks".

docs/components/inputMapper.base.md
```diff
@@ -0,0 +1,282 @@
...
+the web page left. Pressing the thumbstick in opposite direction scrolls the web page right.
+
+### `{inputMapper.base}.scrollVertically(value, speedFactor, invert)`
+
+- `value {Number}` Current value of the gamepad input.
+- `speedFactor {Number}` Times by which the speed of vertical scrolling should be increased.
+- `invert {Boolean}` Whether the vertical scrolling should be in opposite order (see below).
+- Returns: Nothing.
+
+Scrolls the web page in upward and downward direction using gamepad thumbsticks. Calls the
```
the-t-in-rtf, line 278: Another variation, this should just be "scrolls the page upward and downward using the gamepad thumbsticks".

docs/components/inputMapper.base.md
the-t-in-rtf: Much better.  I commented on the two remaining variations that still need to be updated.

docs/components/inputMapper.base.md
dmahajan980: Sorry for putting those backticks on the History API. But I think we should put those on "chrome.storage" as we are using it in the same way, i.e. `chrome.storage.local.get`, `chrome.storage.local.set`, et cetera.

docs/components/inputMapper.base.md
dmahajan980: > But I guess you have tried this with a different namespace and had problems?

@the-t-in-rtf I haven't tried this. But I assumed that Infusion doesn't provide access to sibling namespaces unless the parent namespace is specified. That's the reason I added it.

the-t-in-rtf: It's not needed.  Including the file that defines the required namespaces is sufficient.  To give a node example, say I have a file called `spaceone.js` that defines a namespace and a static function:

```
var fluid = require("infusion");

var spaceone = fluid.registerNamespace("spaceone");

spaceone.sayHello = function () {
    console.log("Hello.");
}
```

If I require that file from another file, I can make use of the first file's functions without ever defining that namespace:

```
var fluid = require("infusion");

require("./spaceone.js");

var spacetwo = fluid.registerNamespace("spacetwo");
fluid.defaults("spacetwo.grade", {
    gradeNames: ["fluid.component"],
    listeners: {
        "onCreate.sayHello": {
            funcName: "spaceone.sayHello"
        }
    }
});

spacetwo.grade();
```

The alternate namespace is visible from the second file and the code works as expected.

docs/components/inputMapper.base.md
the-t-in-rtf: Sounds good.

docs/components/inputMapper.base.md
dmahajan980: @the-t-in-rtf thanks for the example! It does clarify a few things that I misunderstood. Also, I've updated the pull. Kindly take a look over it.

PR #54, opened 2020-08-13
Added documentation for inputMapper component and updated visibilityChangeTracker to pass custom configuration options

dmahajan980:
While writing the `inputMapper` component's documentation, I realized that we are not providing any means to pass in custom configuration options to the component, given how it's handled for multiple tabs and windows. This pull request also addresses that issue.

### DIFF
--- a/docs/components/inputMapper.md
+++ b/docs/components/inputMapper.md
@@ -1,0 +1,216 @@
+<!--
+Copyright (c) 2020 The Gamepad Navigator Authors
+See the AUTHORS.md file at the top-level directory of this distribution and at
+https://github.com/fluid-lab/gamepad-navigator/raw/master/AUTHORS.md.
+
+Licensed under the BSD 3-Clause License. You may not use this file except in
+compliance with this License.
+
+You may obtain a copy of the BSD 3-Clause License at
+https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
+-->
+
+# `gamepad.inputMapper`
+
+<!-- TODO: Add links to the messageListener component's documentation -->
+
+This component extends the [`inputMapper.base`](inputMapper.base.md) grade to provide **inter-web page navigation**
+features. It should be used with the `messageListener` component as the navigation-producing invokers rely on background
+scripts to work correctly.
+
+## Using this grade
+
+The component can be used by creating its instance. However, you might face issues as multiple tabs will read gamepad
+inputs at the same time. To avoid those issues, you should use `gamepad.visibilityChangeTracker` in combination with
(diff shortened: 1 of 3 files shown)


docs/components/inputMapper.md
```diff
@@ -0,0 +1,216 @@
...
+- Returns: Nothing.
+
+Restores the focus on the last focused element after the state of browser tab is changed from **hidden** to **visible**
+or after the web page is reloaded due to refreshing or history navigation.
+
+### `{inputMapper}.updateControls()`
+
+- Returns: Nothing.
+
+Updates the gamepad configuration saved by the user.
```
the-t-in-rtf, line 71: I'm not sure I understand the meaning here.  You should probably expand this to cover both what is happening technically (model variables updated, components recreated, whatever) and practically (i.e. what effect if any is noticeable as a result of triggering this invoker).

docs/components/inputMapper.md
```diff
@@ -0,0 +1,216 @@
...
+Sends a message to the `messageListener` component present in the background script to restore the size of current
+browser window using buttons and triggers.
+
+### `{inputMapper}.thumbstickWindowSize(value, invert)`
+
+- `value {Number}` Current value of the gamepad input.
+- `invert {Boolean}` Whether the thumbstick direction for zoom should be in opposite order (see below).
+- Returns: Nothing.
+
+Sends a message to the `messageListener` component present in the background script to change the size of current
```
the-t-in-rtf, line 204: "the current" is better here.

src/js/content_scripts/input-mapper.js
```diff
@@ -178,19 +183,23 @@ https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
...
+
             /**
              * Create an instance of the inputMapper when the tab/window is focused
              * again and start reading gamepad inputs (if any gamepad is connected).
              */
             if (visibilityStatus === "visible") {
-                // Obtain the saved configuration of the gamepad.
-                inputMapperInstance = gamepad.inputMapper();
+                // Pass the configuration options to the inputMapper component.
+                inputMapperInstance = gamepad.inputMapper(configurationOptions);
```
the-t-in-rtf, line 202: Can you please leave a TODO here suggesting that we refactor to use a dynamic component here?  In general creating components dynamically like this is discouraged.  Using dynamic components would also make this specific issue much easier to address, as you would  have already included any configurable options as IoC references in the definition of the dynamic component.

dmahajan980:
@the-t-in-rtf I've updated the pull. Kindly review it again.

Thanks

PR #55, opened 2020-08-13
Added documentation for messageListener component

dmahajan980:


docs/components/messageListener.md
```diff
@@ -0,0 +1,168 @@
...
+var messageListenerInstanceTwo = my.messageListener.grade();
+```
+
+## Component Options
+
+This component does not provide any configuration options.
+
+## Non-navigation Invokers
+
+### `{messageListener}.addListener()`
```
the-t-in-rtf, line 39: If you are making use of the `addListener` invoker inherited from `fluid.component`, you should mention that and provide a link to that documentation.

docs/components/messageListener.md
```diff
@@ -0,0 +1,168 @@
...
+
+- Returns: Nothing.
+
+Listens to the messages received from the [`inputMapper`](inputMapper.md) component in the content scripts and triggers
+the [actionExecutor](#messagelisteneractionexecutoractiondata) invoker as a callback function.
```
the-t-in-rtf, line 44: I had to refresh my own memory, `addListener` is a method provided by individual events.  I do think it would be better to find a name that is not already widely used in other parts of Infusion.

docs/components/messageListener.md
```diff
@@ -0,0 +1,168 @@
...
+Identifies and triggers the **navigation-producing invokers** according to the `actionData`. The navigation-producing
+invoker is triggered only if the component provides it. Each of these invokers accepts the following arguments:
+
+- `tabId`: ID of the currently active tab.
+- `invert`: Whether direction of the given action should be in the opposite order (for thumbsticks).
+- `active`: Whether a new tab or window should be focused when opened.
+- `homepageURL`: URL that a new tab or window should load when opened.
+- `left`: Position of the current browser window from the screen's left edge (in pixels).
+
+The invokers use only those arguments required to perform their specific action.
```
the-t-in-rtf, line 61: You should spell out how these fields look in a sample `actionData` object (and add a "see below" to the description of `actionData`).

docs/components/messageListener.md
```diff
@@ -0,0 +1,168 @@
...
+
+## Navigation Invokers
+
+### `{messageListener}.openNewTab(active, homepageURL)`
+
+- `active {Boolean}` Whether a new browser tab should be active when opened.
+- `homepageURL {String}` URL that a new browser tab should load when opened.
```
the-t-in-rtf, line 68: "The URL"

docs/components/messageListener.md
```diff
@@ -0,0 +1,168 @@
...
+- `homepageURL`: URL that a new tab or window should load when opened.
+- `left`: Position of the current browser window from the screen's left edge (in pixels).
+
+The invokers use only those arguments required to perform their specific action.
+
+## Navigation Invokers
+
+### `{messageListener}.openNewTab(active, homepageURL)`
+
+- `active {Boolean}` Whether a new browser tab should be active when opened.
```
the-t-in-rtf, line 67: You might want to use a word other than "active" to describe this.  I assume you mean "focused"?  If so, changing the description is enough.

docs/components/messageListener.md
```diff
@@ -0,0 +1,168 @@
...
+### `{messageListener}.goToNextTab()`
+
+- Returns: Nothing.
+
+Switches from the current tab to the next tab in the active browser window. If the active tab is the last tab in the
+list, it will switch to the first tab.
+
+### `{messageListener}.openNewWindow(active, homepageURL)`
+
+- `active {Boolean}` Whether a new browser window should be active when opened.
```
the-t-in-rtf, line 98: Same comment here about clarifying what "active" means in the description.

docs/components/messageListener.md
dmahajan980: It's not the `addListener` from `fluid.component`. Rather it uses the one provided by Chrome Extension APIs, i.e.  `chrome.runtime.onMessage.addListener` which listens to the signals from content scripts. But sure, I'll add a link to its documentation.

docs/components/messageListener.md
```diff
@@ -0,0 +1,168 @@
...
+- `homepageURL {String}` URL that a new browser window should load when opened.
```
the-t-in-rtf, line 99: "The URL"

docs/components/messageListener.md
```diff
@@ -0,0 +1,168 @@
...
+
+Restores the size of the currently active browser window. The previous dimensions and positions are taken from the
+`windowProperties` member variable if stored previously. Otherwise, a set of default values are used for the window's
+position and dimensions.
+
+### `{messageListener}.reopenTabOrWindow()`
+
+- Returns: Nothing.
+
+Reopens the last closed browser session using the
```
the-t-in-rtf, line 166: I would use "tab or window" instead of "session" here.

docs/components/messageListener.md
dmahajan980: From this comment and the previous one, I do think it'd be better if we use some other name for the invoker. Maybe should I add a TODO for now?

docs/components/messageListener.md
```diff
@@ -0,0 +1,168 @@
...
+
+Switches from the currently active browser window to the previous browser window. If the active window is the first
+window in the list, it will switch to the last window.
+
+### `{messageListener}.goToNextWindow()`
+
+- Returns: Nothing.
+
+Switches from the currently active browser window to the next browser window. If the active window is the last window in
+the list, it will switch to the first window.
```
the-t-in-rtf, line 125: I think it would be clearer to vary the wording a bit and not always use "switch".  It's kind of the same as my comments on "action", if you use synonyms or otherwise avoid using the same word over and over, you give people more chance to pick up on a word whose meaning would not otherwise be clear.

In this case you also have the problem of "it" not referring to anything described in the paragraph so far.   I would suggest "focus will move" instead of "it will switch" for all variations on this.

PR #56, opened 2020-08-16
Added documentation for configurationPanel component

### LINKED ISSUE #23: Add documentation for all components, especially the input mapper.
The range of supported actions and options for input mappings has come to the point where it would be useful even for reviews to have a bit of documentation about the allowed/expected options, their meaning, and what each component is meant to do.

As an example of the format we have used in previous work, [check out this piece of the fluid-express documentation](https://github.com/fluid-project/fluid-express/blob/master/docs/headerMiddleware.md).  The key things that need to be covered at least briefly are:

1. What the component is meant to do.
2. What options it supports.
3. What invokers it provides, what parameters they accept, and what they return.

The style we have used previously is nowhere near as formal as JSDocs, but is similar in spirit.

dmahajan980:
Resolves #23 

This pull marks the completion of the documentation for all the components in the project.

### DIFF
--- a/docs/components/configurationPanel.md
+++ b/docs/components/configurationPanel.md
@@ -1,0 +1,234 @@
+<!--
+Copyright (c) 2020 The Gamepad Navigator Authors
+See the AUTHORS.md file at the top-level directory of this distribution and at
+https://github.com/fluid-lab/gamepad-navigator/raw/master/AUTHORS.md.
+
+Licensed under the BSD 3-Clause License. You may not use this file except in
+compliance with this License.
+
+You may obtain a copy of the BSD 3-Clause License at
+https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
+-->
+
+# `gamepad.configurationPanel`
+
+This is a [`fluid.viewComponent`](https://tinyurl.com/y4u8xwpd) and creates the configuration options (Action Dropdown,
+Speed Factor Input Box, et cetera) for each gamepad input. This component handles which configuration options should be
+displayed and what list of actions should be available for a given gamepad input. At the same time, the component also
+synchronizes the value of these configuration options with the configuration used by the gamepad and lets you change the
+existing gamepad configuration wither by creating a new gamepad configuration or by switching to the default
+configuration. The new gamepad configuration is saved by the component in the Chrome's storage using the
+[`chrome.storage`](https://developer.chrome.com/extensions/storage) API.
+
+## Using this grade
+
+The component can be used by creating its instance or by making a custom component using the grade.
+
+``` javascript
+// Either create an instance of the configurationPanel component.
+var configurationPanelInstanceOne = gamepad.messageListener(".configuration-dashboard");
+
+// Otherwise create a custom component using the configurationPanel grade.
+fluid.defaults("my.configurationPanel.grade", {
+    gradeNames: ["my.CustomGrade", "gamepad.configurationPanel"]
+});
+var configurationPanelInstanceTwo = my.configurationPanel.grade(".configuration-dashboard");
+```
+
+## Component Options
+
+The following component configuration options are supported:
+
+| Option | Type | Description |
+| :---: | :---: | :--- |
+| `description` | `{Object}` | Contains a list of gamepad inputs for use in **Input Name** dropdown of the configuration panel. The input names are stored in two separate child objects, namely `buttons` and `axes`. |
+| `message` | `{Object}` | Contains the list of actions supported by the selected gamepad in two separate child objects, namely `buttons` and `axes` input. The key name should match with the name of the navigation invoker in the [`inputMapper`](inputMapper.md) component. These actions are used in the **Action** dropdown of the configuration panel. |
+| `actions` | `{Object}` | Specifies whether the selected action supports `speedFactor`, `invert`, or `background` configuration options. |
+
+These options can be provided to the component in the following ways:
+
+``` javascript
+// Either pass options inside the object as an argument while creating an instance of configurationPanel component.
+var configurationPanelInstanceOne = gamepad.configurationPanel({
+    description: {
+        buttons: { "16": "Button 16: Badge Icon Button" },
+        axes: { "0": "Left Thumbstick Horizontal Direction" }
+    },
+    message: {
+        buttons: {
+            null: "None",
+            scrollLeft: "Scroll left",
+            scrollRight: "Scroll right"
+        },
+        axes: {
+            null: "None",
+            thumbstickTabbing: "Focus on the previous/next element"
+        }
+    },
+    actions: {
+        speedFactorOption: ["scrollLeft", "scrollRight", "thumbstickTabbing"],
+        invertOption: ["thumbstickTabbing"]
+    }
+});
+
+// Otherwise pass it as default options in a custom component using the configurationPanel grade.
+fluid.defaults("my.configurationPanel.grade", {
+    gradeNames: ["gamepad.configurationPanel"],
+    cutoffValue: 0.10,
+    scrollInputMultiplier: 25
+});
+var configurationPanelInstanceTwo = my.configurationPanel.grade({
+    description: {
+        buttons: { "16": "Button 16: Badge Icon Button" },
+        axes: { "0": "Left Thumbstick Horizontal Direction" }
+    },
+    message: {
+        buttons: {
+            null: "None",
+            scrollLeft: "Scroll left",
+            scrollRight: "Scroll right"
+        },
+        axes: {
+            null: "None",
+            thumbstickTabbing: "Focus on the previous/next element"
+        }
+    },
+    actions: {
+        speedFactorOption: ["scrollLeft", "scrollRight", "thumbstickTabbing"],
+        invertOption: ["thumbstickTabbing"]
+    }
+});
+```
+
+## Invokers
+
+### `{configurationPanel}.createMenu()`
+
+- Returns: Nothing.
+
+Creates a separate configuration menu for each gamepad input, as mentioned in the `description` option and is triggered
+when the component is created. This invoker calls other invokers to create the configuration options (Action dropdown,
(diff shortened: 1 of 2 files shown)


docs/components/configurationPanel.md
```diff
@@ -0,0 +1,224 @@
...
+Licensed under the BSD 3-Clause License. You may not use this file except in
+compliance with this License.
+
+You may obtain a copy of the BSD 3-Clause License at
+https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
+-->
+
+# `gamepad.configurationPanel`
+
+This is a `viewComponent` and handles the state and functionality of the configuration panel. It also creates the
```
the-t-in-rtf, line 15: I'd prefer the full gradeName, i.e. `fluid.viewComponent`. Please link to the docs as well.

the-t-in-rtf: "state and functionality" is not very clear, nor is "creates the configuration options".  I'd suggest reworking this whole paragraph.

docs/components/configurationPanel.md
```diff
@@ -0,0 +1,224 @@
...
+```
+
+## Invokers
+
+### `{configurationPanel}.createMenu()`
+
+- Returns: Nothing.
+
+Creates a separate configuration menu for each gamepad input, as mentioned in the `description` option and is triggered
+when the component is created. This invoker calls other invokers to create the configuration options and to listen to
```
the-t-in-rtf, line 106: I still am not clear on what you mean by "create the conifguration options".

docs/components/configurationPanel.md
```diff
@@ -0,0 +1,224 @@
...
+various interaction-based events and changes on the configuration panel.
```
the-t-in-rtf, line 107: "to the configuration panel" sounds better

docs/components/configurationPanel.md
```diff
@@ -0,0 +1,224 @@
...
+
+### `{configurationPanel}.createInputActionDropdown(inputIdentifier, configMenu, inputType, currentValue)`
+
+- `inputIdentifier {String}` A description of the gamepad input type and its index. For example, "button-0".
```
the-t-in-rtf, line 111: Sorry for not catching this earlier,  but it seems less than ideal to in essence encode two pieces of data in a string like this using your own convention.  I'd suggest filing a ticket to cover converting this to use the same dot-notation supported by `fluid.get` and `fluid.put`, i.e. `button.0` instead of `button-0`.

You would replace whatever is parsing the string into the two pieces of data (or combining it) with `fluid.get` or `fluid.set` rather than string splitting and concatenation.

docs/components/configurationPanel.md
```diff
@@ -0,0 +1,224 @@
...
+- `configMenu {Object}` The DOM Object used for storing the configuration options for a given gamepad input.
+- `inputType {Boolean}` The type of the gamepad input, i.e. "buttons" or "axes".
+- `currentValue {Object}` The value of the currently selected action for the given gamepad input.
```
the-t-in-rtf, line 114: If there's only one type of "value" that gets stored, it would be better to rename this to reflect what the value is, for example `currentAction`.

docs/components/configurationPanel.md
```diff
@@ -0,0 +1,224 @@
...
+- Returns: Nothing.
+
+Creates a separate configuration menu for each gamepad input, as mentioned in the `description` option and is triggered
+when the component is created. This invoker calls other invokers to create the configuration options and to listen to
+various interaction-based events and changes on the configuration panel.
+
+### `{configurationPanel}.createInputActionDropdown(inputIdentifier, configMenu, inputType, currentValue)`
+
+- `inputIdentifier {String}` A description of the gamepad input type and its index. For example, "button-0".
+- `configMenu {Object}` The DOM Object used for storing the configuration options for a given gamepad input.
```
the-t-in-rtf, line 112: You should always note (here and in the JSDocs) when. you directly modify something that is passed to you as an argument.

docs/components/configurationPanel.md
```diff
@@ -0,0 +1,224 @@
...
+- Returns: Nothing.
+
+Creates an "Action" label and dropdown for the given gamepad input and then injects these elements into the `configMenu`
+object passed as an argument. The attributes and classnames of these elements are also configured before they are
+injected, and the value of the "Action" dropdown is set equal to the currently selected action.
+
+### `{configurationPanel}.createSpeedFactorOption(inputIdentifier, configMenu, currentValue)`
+
+- `inputIdentifier {String}` A description of the gamepad input type and its index. For example, "button-0".
+- `configMenu {Object}` The DOM Object used for storing the configuration options for a given gamepad input.
```
the-t-in-rtf, line 124: Same comment here about the fact that the argument will be modified in place.

docs/components/configurationPanel.md
```diff
@@ -0,0 +1,224 @@
...
+- Returns: Nothing.
+
+Creates a "Speed Factor" label and input field for the given gamepad input and then injects these elements into the
+`configMenu` object passed as an argument. The attributes and classnames of these elements are also configured before
+they are injected and the value of the "Speed Factor" input field is set equal to the current `speedFactor` value.
+
+### `{configurationPanel}.createCheckbox(inputIdentifier, configMenu, isAxes, currentValue)`
+
+- `inputIdentifier {String}` A description of the gamepad input type and its index. For example, "button-0".
+- `configMenu {Object}` The DOM Object used for storing the configuration options for a given gamepad input.
```
the-t-in-rtf, line 135: Same comment here about documenting the fact that you modify the argument passed to you.

docs/components/configurationPanel.md
```diff
@@ -0,0 +1,224 @@
...
+
+Toggles the state of "Discard Changes" and "Save Changes" button when the value of any configuration option on the panel
+is changed. The new value of the configuration options are compared to the last saved values, and if those are
+unchanged, then the buttons will be disabled. Otherwise, the buttons are enabled.
+
+### `{configurationPanel}.storeUnsavedChanges()`
+
+- Returns: Nothing.
+
+Stores all the unsaved changes in Chrome's storage when the value of any configuration option is changed. These values
```
the-t-in-rtf, line 222: Please link to the Chrome storage API docs at least the first time you mention it on a page.

docs/components/configurationPanel.md
dmahajan980: Thanks for pointing this out! I've filed the ticket [here](https://github.com/fluid-lab/gamepad-navigator/issues/58).

dmahajan980:
@the-t-in-rtf thanks for the feedback. I've updated the pull. Kindly take a look over it.

docs/components/configurationPanel.md
```diff
@@ -12,9 +12,13 @@ https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
 
 # `gamepad.configurationPanel`
 
-This is a `viewComponent` and handles the state and functionality of the configuration panel. It also creates the
-configuration options and stores the custom gamepad configuration for use in the [`inputMapper`](inputMapper.md)
-component.
+This is a [`fluid.viewComponent`](https://tinyurl.com/y4u8xwpd) and creates the configuration options (Action Dropdown,
```
the-t-in-rtf, line 15: "and creates" is odd, but rather than fix that, I'd suggest simplifying to "This `fluid.viewComponent` creates".

docs/components/configurationPanel.md
```diff
@@ -12,9 +12,13 @@ https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
...
+Speed Factor Input Box, et cetera) for each gamepad input. This component handles which configuration options should be
+displayed and what list of actions should be available for a given gamepad input. At the same time, the component also
+synchronizes the value of these configuration options with the configuration used by the gamepad and lets you change the
+existing gamepad configuration wither by creating a new gamepad configuration or by switching to the default
```
the-t-in-rtf, line 19: Should be "either".

docs/components/configurationPanel.md
```diff
@@ -12,9 +12,13 @@ https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
...
+configuration. The new gamepad configuration is saved by the component in the Chrome's storage using the
```
the-t-in-rtf, line 20: you can just leave out "in the Chrome's storage".

docs/components/configurationPanel.md
```diff
@@ -219,6 +227,8 @@ unchanged, then the buttons will be disabled. Otherwise, the buttons are enabled
 
 - Returns: Nothing.
 
-Stores all the unsaved changes in Chrome's storage when the value of any configuration option is changed. These values
-are saved separately without applying the gamepad configuration and are used to prevent loss of unsaved configuration
-due to the panel getting closed accidentally. These values are restored when the panel is reopened.
+Stores all the unsaved changes in Chrome's storage (using
```
the-t-in-rtf, line 230: You can leave out "in Chrome's storage" and move the other part out of the parentheses.

docs/components/configurationPanel.md
```diff
@@ -219,6 +227,8 @@ unchanged, then the buttons will be disabled. Otherwise, the buttons are enabled
...
+[`chrome.storage`](https://developer.chrome.com/extensions/storage) API) when the value of any configuration option is
+changed. These values are saved separately without applying the gamepad configuration and are used to prevent loss of
```
the-t-in-rtf, line 232: I'd avoid using "saved" for quite so many things here, i.e. you might also mean saving unsaved changes rather than preserving the values in the model.  Something like "Unsaved changes are preserved to prevent losing them when the configuration panel is closed" sounds better.

docs/components/configurationPanel.md
dmahajan980: Sorry, that was a typing error. Thanks for pointing this out!

dmahajan980:
@the-t-in-rtf I've updated the pull. Kindly review it again.

Thanks

PR #65, opened 2020-08-20
Added the extension icon, grunt tasks, and steps to publish the extension

### LINKED ISSUE #63: Publish the extension on the Chrome Web Store
The project has reached to the point where the initial list of features is more or less complete along with their documentation. It is ready to be published on the Chrome Web Store so that the other users can use it and leave suggestions and feedback for further improvements.

The following tasks need to be completed before publishing the extension:
- Add an icon for the extension.
- Add steps to publish the extension in the README file.

dmahajan980:
Resolves #63 

The following tasks have been completed:
- [x] Add an icon for the extension.
- [x] Add steps to publish the extension in the README file.

After this pull request gets merged, we can initiate the process of publishing the extension.


README.md
```diff
@@ -146,3 +146,61 @@ You can check out the recorded demo to see how the navigator works.
...
+   3. Ensure that the code in the `master` branch is working as expected.
+      1. Run tests: `npm run test`
+      2. Lint: `grunt lint`
+      3. Manual test build.
+         1. Create a build and load the generated unpacked extension into Chrome.  
+            (Refer to the [Installation](#installation) section for more details)
+         2. Test all of the available configuration options (Action, Speed Factor, et cetera) and ensure that they work
+            on the browser correctly.
+            1. Refresh any browser tabs/windows that were open before installing the extension.
+            2. The actions and the associated configuration options should be tested individually and in combinations
```
the-t-in-rtf, line 165: "combination" is better here.

README.md
```diff
@@ -146,3 +146,61 @@ You can check out the recorded demo to see how the navigator works.
...
+               to ensure that they are working correctly. For example, testing **Scroll Vertically** separately, then
+               changing its **Speed Factor**, and selecting the **Invert Action** option.
+            3. Test the same action and the associated configuration options with at least one different button,
+               thumbstick, and trigger.
+            4. Multiple web pages should be tested to ensure that everything works correctly.
+
+2. Create the release package.
+   1. Create a release build: `grunt build`
+   2. Create a zip archive of the contents of the `dist` directory. The zip file will be uploaded to the
```
the-t-in-rtf, line 174: The creation of the zip should probably be a new Grunt task or an npm script.  That way there's less risk of typos or other human errors in the process.

README.md
```diff
@@ -146,3 +146,61 @@ You can check out the recorded demo to see how the navigator works.
...
+      [Chrome Web Store](https://chrome.google.com/webstore/category/extensions).
+
+3. Publish to the [Chrome Web Store](https://chrome.google.com/webstore/category/extensions) and create a GitHub
```
the-t-in-rtf, line 177: I would move "create a GitHub release" to a fourth point rather than combining it with the Chrome Web Store release.

the-t-in-rtf: Also, I just noticed that the details about publishing a GitHub release appear between sections on publishing to the Chrome Web Store and verifying that work. I'd suggest reorganising so that those two sections on the Chrome Web Store appear one after the other.  For example, moving "Create a GitHub Release" to the end.

README.md
```diff
@@ -146,3 +146,61 @@ You can check out the recorded demo to see how the navigator works.
...
+   release.
+   1. Go to the Developer Dashboard on the Chrome Web Store and log in.
+   2. On the Developer Dashboard, click "Edit" (located on the Gamepad Navigator's right-hand side).
+   3. On the Gamepad Navigator edit page, click "Upload updated package" and upload the zip created in step 2.2 above.
+   4. Update the "Detailed description" field as necessary. Similar information is contained in this README.
+   5. Update the screenshots and videos if necessary. They will need to be the exact size requested.
```
the-t-in-rtf, line 183: You should provide guidance on how to easily determine "if necessary", for example, looking at the list of change files in the release and confirming whether the icons have been updated in a given release.

README.md
```diff
@@ -146,3 +146,61 @@ You can check out the recorded demo to see how the navigator works.
...
+         reference.
+      5. Attach the build zip file created in step 2.2. Before uploading, make sure the file is named
+         "build_{tag_name}.zip". For example, build_v0.1.0-beta.zip.
+      6. If the release is a beta version, check the "This is a pre-release" option.
+      7. After all the information has been entered correctly, click on the "Publish release" button.
+
+4. Verify the published Gamepad Navigator Chrome extension.
+   1. Ensure that the contents of the Gamepad Navigator on the Chrome Web Store appear correct. Double-check the
+      details like version number/name, descriptions, screenshots, et cetera.
+   2. Install the version from the Chrome Web Store, and run through the manual testing again. (See step 1.2.3 above)
```
the-t-in-rtf, line 204: I know what you mean, but "the version" could be replaced with "the extension" here.

README.md
```diff
@@ -146,3 +146,61 @@ You can check out the recorded demo to see how the navigator works.
...
+   3. If everything is working, announce the release where required (for example, the fluid-work mailing list). If
+      there are any issues, fix them and repeat the process.
```
the-t-in-rtf, line 206: I would move the second sentence to point 3 and have the announcements be a clear step 4 that you hit if there are no problems.

README.md
```diff
@@ -146,3 +146,61 @@ You can check out the recorded demo to see how the navigator works.
...
+      1. The actual publishing to the Chrome Web Store will take some time and may need to go through a review process.
+   8. Tag the master branch with the release. For example, v1.1.0-beta.
+   9. Create a GitHub release for the tag.
+      1. Go to the [Gamepad Navigator](https://github.com/fluid-lab/gamepad-navigator) GitHub page.
+      2. Click on "Create a new release" under the "Releases" section on the web page's right-hand side.
+      3. For the "Tag Version" and "Release Title", enter the tag name created in step 3.8 (for example, v1.1.0-beta).
+      4. For the description, add a summary of changes and any other relevant information. View prior releases for your
+         reference.
+      5. Attach the build zip file created in step 2.2. Before uploading, make sure the file is named
+         "build_{tag_name}.zip". For example, build_v0.1.0-beta.zip.
```
the-t-in-rtf, line 197: Your use of `-beta` here suggests to me that you should have two Grunt tasks (or npm scripts), one for a dev release, one for a full release.

README.md
```diff
@@ -146,3 +146,61 @@ You can check out the recorded demo to see how the navigator works.
...
+   2. On the Developer Dashboard, click "Edit" (located on the Gamepad Navigator's right-hand side).
+   3. On the Gamepad Navigator edit page, click "Upload updated package" and upload the zip created in step 2.2 above.
+   4. Update the "Detailed description" field as necessary. Similar information is contained in this README.
+   5. Update the screenshots and videos if necessary. They will need to be the exact size requested.
+   6. Click "Preview Changes".
+      1. Verify that everything appears correct. Pay particular attention to anything that was changed. For example,
+         version number/name, description, screenshots, et cetera.
+   7. If everything appears correct, publish the changes.
+      1. The actual publishing to the Chrome Web Store will take some time and may need to go through a review process.
+   8. Tag the master branch with the release. For example, v1.1.0-beta.
```
the-t-in-rtf, line 189: I can also see making a Grunt task or npm script to help with the tagging, especially since you'll want to do whatever you can to make sure the tag lines up between the published zip and the tag on GitHub.

src/manifest.json
```diff
@@ -31,7 +31,14 @@
...
         ]
     },
+    "icons": {
+        "16": "images/gamepad-icon-16px.png",
+        "32": "images/gamepad-icon-32px.png",
+        "48": "images/gamepad-icon-48px.png",
+        "128": "images/gamepad-icon-128px.png"
+    },
     "browser_action": {
+        "default_icon": "images/gamepad-icon.png",
```
the-t-in-rtf, line 41: Why did you use `./html` for the popup but not `./images` here?  They should be consistent unless there's a reason they need to be different.

README.md
the-t-in-rtf, line 204: See if you can find a better way to link to a step, it's less brittle.  As an example, this is already out of day, that bullet point is now `1.3.3`.

README.md
```diff
@@ -146,3 +146,61 @@ You can check out the recorded demo to see how the navigator works.
...
+   9. Create a GitHub release for the tag.
+      1. Go to the [Gamepad Navigator](https://github.com/fluid-lab/gamepad-navigator) GitHub page.
+      2. Click on "Create a new release" under the "Releases" section on the web page's right-hand side.
+      3. For the "Tag Version" and "Release Title", enter the tag name created in step 3.8 (for example, v1.1.0-beta).
```
the-t-in-rtf, line 193: I think the previous step is so close to this one that you can just say "above" instead of `in step 3.8`.

README.md
```diff
@@ -146,3 +146,61 @@ You can check out the recorded demo to see how the navigator works.
...
+      4. For the description, add a summary of changes and any other relevant information. View prior releases for your
+         reference.
+      5. Attach the build zip file created in step 2.2. Before uploading, make sure the file is named
```
the-t-in-rtf, line 196: While this reference to the bullet point step is correct, it would be safer to also find a way to link directly to it.  One way is to use headings and sub-headings instead of a bulleted list.  But look around, maybe there's a way to directly link between bullet points.

the-t-in-rtf: It's fine for now.  Eventually you may want to split off sub-checklists, for now I'm happy as long as the references to bullet numbers match.

Gruntfile.js
```diff
@@ -73,16 +80,270 @@ module.exports = function (grunt) {
...
+            })
+            .catch(function (errorMessage) {    // eslint-disable-line
+                grunt.log.error(errorMessage);
+
+                // Indicate that the task has completed due to failure.
+                done();
+            });
+
+        // Creates a release in the repository.
+        var createRelease = function (callback, taskOptions) {   // eslint-disable-line
```
the-t-in-rtf, line 238: Why did you have to disable ESLint here?

README.md
```diff
@@ -137,6 +137,8 @@ The default action for each gamepad input is as follows.<br>
 | `Right Thumbstick Horizontal Direction` | History navigation | - | `false` |
 | `Right Thumbstick Vertical Direction` | Focus on the previous/next element | `2.5` | `false` |
 
+// TODO: Mention the keyboard shortcut to open the configuration panel.
```
the-t-in-rtf, line 140: Wrong type of comment here, this is visible in the rendered page.

README.md
```diff
@@ -150,57 +152,68 @@ You can check out the recorded demo to see how the navigator works.
...
+         1. Dev release with MAJOR version updates: `grunt updateVersion:dev`
+         2. Dev release with MINOR version updates: `grunt updateVersion:minor:dev`
+         3. Dev release with PATCH version updates: `grunt updateVersion:patch:dev`
+      4. Push the changes to the `master` branch.
    2. Ensure that all of the code that should be published has been merged into the `master` branch.
    3. Ensure that the code in the `master` branch is working as expected.
-      1. Run tests: `npm run test`
-      2. Lint: `grunt lint`
+      1. Lint: `grunt lint`
+      2. Run tests: `npm run test`
```
the-t-in-rtf, line 171: `npm test` is fine.

README.md
```diff
@@ -150,57 +152,68 @@ You can check out the recorded demo to see how the navigator works.
 ## Publishing to the Chrome Web Store
 
 1. Prepare the code and the `master` branch.
-   1. Increase the `version` number in the [manifest](src/manifest.json#L6) file and push changes to the `master`
-      branch.
+   1. Update the `version` number in the [manifest](src/manifest.json#L3) and [package](package.json#L3) files.
+      1. The project follows semantic versioning, i.e. the version is in the format "{MAJOR.MINOR.PATCH}[-dev]". For
+         example, 1.0.15 and 1.1.9-dev.  
+         (Refer to the [semantic versioning](https://semver.org/) standard documentation for more details)
+      2. If the release is a **full release**, use one of the following commands.
```
the-t-in-rtf, line 159: I have tested the release updating, in general it's very nice.  I would argue that the default should not be a full major release.  Most often you will be creating dev patch releases, IMO that should be the default.  You should have to indicate that the release is "full".

Is it possible to use flags rather than using multiple colons in the task name?  That seems preferable.

README.md
```diff
@@ -150,57 +152,68 @@ You can check out the recorded demo to see how the navigator works.
 ## Publishing to the Chrome Web Store
```
the-t-in-rtf, line 152: Looking at the page altogether, it's getting a bit long.  The home page serves in part as a marketing page for your work, and for that purpose, shorter is better.   I'd suggest moving the publishing content to a separate page under `docs` and linking from the README.

the-t-in-rtf:
@dmahajan980, I left a few small comments.  I didn't test the publishing or GitHub steps.  I would suggest doing that together as follows:

1. You create a "dev" release.
2. You publish to the Chrome store and GitHub using your steps.
3. I look through the results of both and give you feedback.
4. When all other work is done, we merge this pull and you cut/publish the full release.

Gruntfile.js
the-t-in-rtf: If there is a good rationale, then at a minimum I would suggest disabling the specific rule with a comment nearby.

dmahajan980: Sorry, that was added by mistake. It can run absolutely fine without the rule. I'll remove it in the next commit.

src/manifest.json
the-t-in-rtf: This still needs to be done apparently.

the-t-in-rtf:
The flags work very well, and do well with switching between dev and live releases.  One thing that needs more thought is the transition from making dev releases to live.  Right now you assume someone is trying to make a patch release by default, so that if I finish testing a dev release and want to switch to a standard release, I end up incrementing the version.  There should be some means of testing `1.0.1-dev` and then releasing `1.0.1`.  Perhaps a check to see if the previous version was a dev release.  If no major/minor/patch flags are set, then simply regenerate the version without the `-dev`.

I do also want you to go through the process using your tools and documentation to publish a dev release, I understand that you need input from @jobara before you can do that.

PR #74, opened 2020-10-06
NOGH: Added Chrome Web Store link and configuration panel keyboard shortcut to README

dmahajan980:
As the Gamepad Navigator has been released earlier on the Chrome Web Store, I think it should be good to add a reference to the Web Store listing somewhere in the README. This pull request adds a Web Store badge referring to the extension on the store. Along with that, the pull request also adds the keyboard shortcut (for opening the configuration panel) to the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,14 @@ reconfigured from the application to redefine what each d-pad, thumbstick, butto
 
 ## Installation
 
+The Gamepad Navigator Chrome Extension is available for free on the
+[Chrome Web Store](https://chrome.google.com/webstore/detail/gamepad-navigator/egilmijcknfacjjbchcacijkknbkgfnd).
+
+[![Chrome Web Store](https://developer.chrome.com/webstore/images/ChromeWebStore_BadgeWBorder_v2_206x58.png)](https://chrome.google.com/webstore/detail/gamepad-navigator/egilmijcknfacjjbchcacijkknbkgfnd)
+
+The extension can also be installed on other Chromium-based browsers using the source code.  
+(Refer to the steps mentioned below)
+
 1. Clone or download the repository.
 
 2. Install [grunt-cli](https://gruntjs.com/) globally:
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -28,6 +28,14 @@ reconfigured from the application to redefine what each d-pad, thumbstick, butto
 
 ## Installation
 
+The Gamepad Navigator Chrome Extension is available for free on the
+[Chrome Web Store](https://chrome.google.com/webstore/detail/gamepad-navigator/egilmijcknfacjjbchcacijkknbkgfnd).
+
+[![Chrome Web Store](https://developer.chrome.com/webstore/images/ChromeWebStore_BadgeWBorder_v2_206x58.png)](https://chrome.google.com/webstore/detail/gamepad-navigator/egilmijcknfacjjbchcacijkknbkgfnd)
+
+The extension can also be installed on other Chromium-based browsers using the source code.  
+(Refer to the steps mentioned below)
```
the-t-in-rtf, line 37: This should go before the period, and "see below" in parentheses is enough.

README.md
```diff
@@ -137,7 +145,8 @@ The default action for each gamepad input is as follows.
 | `Right Thumbstick Horizontal Direction` | History navigation | - | `false` |
 | `Right Thumbstick Vertical Direction` | Focus on the previous/next element | `2.5` | `false` |
 
-<!-- TODO: Mention the keyboard shortcut to open the configuration panel. -->
+The configuration panel can be opened by clicking on the Gamepad Navigator pop-up icon. You can also use the keyboard
+shortcut `Ctrl + Shift + 5` to open the configuration panel.
```
the-t-in-rtf, line 149: I just checked out the branch, rebuilt, and installed the plugin.  The hotkey doesn't seem to work for me.

dmahajan980: @the-t-in-rtf thank you for pointing this out. While I'm not completely sure as to why it isn't working, it might be possible that the hotkey is overlapping with some other feature/functionality in the recent versions of Chrome. I tried changing that to `Alt + Shift + G` and the new shortcut seems to work, at least in Ubuntu (Linux). Could you please try the new shortcut once and confirm whether it works in macOS as well? We can try other solutions and dive deeper into the problem if this doesn't work.

README.md
dmahajan980: Thank you. I've updated the statement as you suggested. Kindly take a look.

README.md
the-t-in-rtf: `Alt + Shift + G` works.  I couldn't figure out why the other hotkey didn't.

PR #75, opened 2020-10-27
NOGH: Added Privacy Policy

dmahajan980:
This pull request adds the Privacy Policy to the project. Once this gets merged, we shall create a new GitHub release addressing the changes in #74, update the extension on the Chrome Web Store, and link the Privacy Policy to the [Fluid Wiki Page](http://wiki.fluidproject.org/).

### DIFF
--- a/PRIVACY_POLICY.md
+++ b/PRIVACY_POLICY.md
@@ -15,24 +15,20 @@ https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
 The Gamepad Navigator allows the user to navigate Chromium-based browsers and web pages using a game controller. It
 provides various features that allow navigation using gamepad such as web page scrolling, tab navigation, click on
 buttons and hyperlinks, et cetera. Due to the nature of browser and web page navigation, the Gamepad Navigator requires
-permission to interact with the content in the active web pages, browser windows, and tabs.
+permission to interact with the content in active web pages, browser windows, and tabs.
 
 ## Information the Gamepad Navigator collects
 
-The Gamepad Navigator provides some web page navigation features that involve interaction with the web page's content.
-To perform certain navigation features (for example, tab navigation) correctly, the elements on the web page may be
-parsed. In addition, the Gamepad Navigator may use the custom gamepad configuration map provided by the user using the
-configuration panel.
+The Gamepad Navigator makes use of user preferences entered using the extension configuration panel. To enable
+navigation features, the content of active websites is also parsed.
 
 ## How the Gamepad Navigator uses the information
 
-The Gamepad Navigator is directed by the user's game controller to perform browser and web page navigation according to
-the configuration map set by default or as provided by the user. Some of the navigation features require access to the
-elements on the web page and perform relevant operations, as needed, For example, "tab navigation" requires filtering
-and sorting the tabbable/clickable elements on the web page prior to performing the action.
+User preferences are used to determine which gamepad inputs activate browser actions. Page content is parsed to
+determine which focusable elements are available, so that the user can navigate between focusable elements as they would
+with the tab key on a keyboard.
 
 ## What information does the Gamepad Navigator share
 
-The Gamepad Navigator does not explicitly share any information. The gamepad configuration map provided using the
-configuration panel is stored locally within the extension for persistence across multiple browser sessions. Elements
-parsed from the web pages are only stored temporarily.
+User preferences entered are stored locally on the user's computer and not shared with any site. Page content used for
+navigation is only retained temporarily in memory, and is not shared with any site.
--- a/PRIVACY_POLICY.md
+++ b/PRIVACY_POLICY.md
@@ -19,8 +19,8 @@ permission to interact with the content in active web pages, browser windows, an
 
 ## Information the Gamepad Navigator collects
 
-The Gamepad Navigator makes use of user preferences entered using the extension configuration panel. To enable
-navigation features, the content of active websites is also parsed.
+The Gamepad Navigator makes use of user preferences entered using the extension configuration panel, which can be
+removed/deleted using panel buttons. To enable navigation features, the content of active websites is also parsed.
 
 ## How the Gamepad Navigator uses the information
 
--- a/PRIVACY_POLICY.md
+++ b/PRIVACY_POLICY.md
@@ -1,0 +1,38 @@
+<!--
+Copyright (c) 2020 The Gamepad Navigator Authors
+See the AUTHORS.md file at the top-level directory of this distribution and at
+https://github.com/fluid-lab/gamepad-navigator/raw/master/AUTHORS.md.
+
+Licensed under the BSD 3-Clause License. You may not use this file except in
+compliance with this License.
+
+You may obtain a copy of the BSD 3-Clause License at
+https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
+-->
+
+# Privacy Policy
+
+The Gamepad Navigator allows the user to navigate Chromium-based browsers and web pages using a game controller. It
+provides various features that allow navigation using gamepad such as web page scrolling, tab navigation, click on
+buttons and hyperlinks, et cetera. Due to the nature of browser and web page navigation, the Gamepad Navigator requires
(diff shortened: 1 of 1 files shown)


PRIVACY_POLICY.md
```diff
@@ -0,0 +1,38 @@
...
+You may obtain a copy of the BSD 3-Clause License at
+https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
+-->
+
+# Privacy Policy
+
+The Gamepad Navigator allows the user to navigate Chromium-based browsers and web pages using a game controller. It
+provides various features that allow navigation using gamepad such as web page scrolling, tab navigation, click on
+buttons and hyperlinks, et cetera. Due to the nature of browser and web page navigation, the Gamepad Navigator requires
+permission to interact with the content in the active web pages, browser windows, and tabs.
```
the-t-in-rtf, line 18: You can remove "the" before "active".

PRIVACY_POLICY.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Information the Gamepad Navigator collects
```
the-t-in-rtf, line 20: What do their guidelines say about "collects"?  I'd like to see how they distinguish between the information you have access to and whether you store or transmit it.

PRIVACY_POLICY.md
```diff
@@ -0,0 +1,38 @@
...
+The Gamepad Navigator is directed by the user's game controller to perform browser and web page navigation according to
+the configuration map set by default or as provided by the user. Some of the navigation features require access to the
+elements on the web page and perform relevant operations, as needed, For example, "tab navigation" requires filtering
+and sorting the tabbable/clickable elements on the web page prior to performing the action.
+
+## What information does the Gamepad Navigator share
+
+The Gamepad Navigator does not explicitly share any information. The gamepad configuration map provided using the
+configuration panel is stored locally within the extension for persistence across multiple browser sessions. Elements
+parsed from the web pages are only stored temporarily.
```
the-t-in-rtf, line 38: you can omit "the" before "web".

PRIVACY_POLICY.md
```diff
@@ -0,0 +1,38 @@
...
+## How the Gamepad Navigator uses the information
+
+The Gamepad Navigator is directed by the user's game controller to perform browser and web page navigation according to
+the configuration map set by default or as provided by the user. Some of the navigation features require access to the
+elements on the web page and perform relevant operations, as needed, For example, "tab navigation" requires filtering
+and sorting the tabbable/clickable elements on the web page prior to performing the action.
+
+## What information does the Gamepad Navigator share
+
+The Gamepad Navigator does not explicitly share any information. The gamepad configuration map provided using the
```
the-t-in-rtf, line 36: Why "explicitly"?  You don't share information at all, right?  I would also spell out "information" even if you end up repeating yourself a little.

PRIVACY_POLICY.md
```diff
@@ -0,0 +1,38 @@
...
+The Gamepad Navigator provides some web page navigation features that involve interaction with the web page's content.
+To perform certain navigation features (for example, tab navigation) correctly, the elements on the web page may be
+parsed. In addition, the Gamepad Navigator may use the custom gamepad configuration map provided by the user using the
+configuration panel.
+
+## How the Gamepad Navigator uses the information
+
+The Gamepad Navigator is directed by the user's game controller to perform browser and web page navigation according to
+the configuration map set by default or as provided by the user. Some of the navigation features require access to the
+elements on the web page and perform relevant operations, as needed, For example, "tab navigation" requires filtering
```
the-t-in-rtf, line 31: I think "to perform" reads better.

the-t-in-rtf:
I general this is OK, I think the wording could be simplified.  I'd also like to see one or more working examples of a privacy policy within the community to make sure we're not missing the intent of words like "collect" and "share".

PRIVACY_POLICY.md
dmahajan980: In general, the Chrome extension [privacy policy FAQ](https://developer.chrome.com/webstore/user_data) has stated a few cases about the information collection or "information handling", to be precise. You can read more about those cases [here](https://developer.chrome.com/webstore/user_data#ques_2). 

> Clipping or scraping content from a website that the user visits, such as taking screenshots or capturing data from a web page

This is one of the cases about information collection/handling that I copied from their FAQ, which comes close to what the Navigator does (even though the data is stored temporarily).

PRIVACY_POLICY.md
dmahajan980: Yes, the extension does not share any information. I just added "explicitly" so that there's no room for confusion for the reader. However, I can re-word the statement if you want.

dmahajan980:
@the-t-in-rtf thanks for reviewing the pull request. As far as the structure of the policy document is concerned, the Chrome extension FAQs have stated that [here](https://developer.chrome.com/webstore/user_data#ques_6). I've used the UIO+ Privacy Policy as a reference. You can read more about that [here](https://github.com/fluid-project/uio-plus/blob/main/PRIVACY_POLICY.md).

PRIVACY_POLICY.md
the-t-in-rtf: Yes, at issue here is whether we "collect" anything at all or "handle" it, and their page is also vague on that.  Their examples do make it sound like the key concerns are around transmitting personal data of any kind.

the-t-in-rtf:
I think we could aim for the level of clarity and brevity of the UIO+ privacy policy, and since we do even less with page content, we have to announce even less.  I'll comment with specifics.

PRIVACY_POLICY.md
the-t-in-rtf: This block is fine now.

PRIVACY_POLICY.md
```diff
@@ -0,0 +1,38 @@
...
+buttons and hyperlinks, et cetera. Due to the nature of browser and web page navigation, the Gamepad Navigator requires
+permission to interact with the content in the active web pages, browser windows, and tabs.
+
+## Information the Gamepad Navigator collects
+
+The Gamepad Navigator provides some web page navigation features that involve interaction with the web page's content.
+To perform certain navigation features (for example, tab navigation) correctly, the elements on the web page may be
+parsed. In addition, the Gamepad Navigator may use the custom gamepad configuration map provided by the user using the
+configuration panel.
+
```
the-t-in-rtf, line 26: I like the UIO+ wording, "makes use of", and I also think you can avoid repeating yourself a bit.  Something like:

```
The Gamepad Navigator makes use of user preferences entered using the extension configuration panel.  To enable navigation features, the content of active websites is also parsed.
```

PRIVACY_POLICY.md
the-t-in-rtf: How about something like:

> User preferences entered are stored locally on the user's computer and not shared with any site.  Page content used for navigation is only retained temporarily in memory, and is not shared with any site.

PRIVACY_POLICY.md
the-t-in-rtf: How about this?

> User preferences entered are stored locally on the user's computer and not shared with any site.  Page content used for navigation is only retained temporarily in memory, and is not shared with any site.

the-t-in-rtf:
Thanks for the clear examples and guidelines, that helped a lot.  I have commented with updated wording suggestions.

dmahajan980:
@the-t-in-rtf, thanks for going through the links and examples and suggesting changes accordingly. I think it'll be a good time to have @jobara review this pull request and then merge it if everything seems okay.

the-t-in-rtf:
I'm happy with the updated wording.  @jobara, did you have any further concerns / comments?

jobara:
> I'm happy with the updated wording. @jobara, did you have any further concerns / comments?

I think it's okay. If there is a way to remove user preferences, it might be worth mentioning that and linking to any docs that explain how.

dmahajan980:
@jobara, thanks for the suggestion. There are a few buttons on the configuration panel that the user can use to remove preferences, and I've updated the policy doc accordingly. Could you and @the-t-in-rtf please review the pull request once? Kindly let me know if you want any further changes.

PR #80, opened 2020-12-20
Updated User Interface and game controller svg

SHIVAJI07:


duhrer:
Apologies for the extreme delay in responding.  I'll be redoing all the images and removing the one in the current configuration shortly.  Unfortunately this means I won't be able to use what you have submitted here.

If you would like to follow/contribute replacement icons, check out #117.

Issue #86, opened 2022-02-15
Q1 2022 Dependency Update


the-t-in-rtf:
Before working on any new features or bug fixes, we should take a minute to update all dependencies as far as possible.

PR #87, opened 2022-02-15
GH-86: Q1 2022 Dependency updates.

the-t-in-rtf:
See #86 for details.


the-t-in-rtf:
We cannot update `node-fetch` to 3.x because it's now an ES6 module.

the-t-in-rtf:
We don't need to upload an updated chrome extension with just these updates, we should do it when we have a new feature.

PR #89, opened 2022-02-15
GH-88: Added npm-audit-resolve check to CI (resolves #88).

### LINKED ISSUE #88: Add tailored security audits to CI builds.
We should fail builds if we update to a dependency that brings in vulnerabilities.  In other projects I have used npm-audit-resolve for this purpose, since it lets you exclude vulnerabilities you choose (such as those brought in by build and test tools).

the-t-in-rtf:
See #88 for details.


the-t-in-rtf:
Had to exclude the auto-generated audit-resolve.json file from linting checks.

PR #92, opened 2022-04-07
GH-91: Make 'click' action better simulate a mouse click and release (resolves #91).

### LINKED ISSUE #91: Make click action fire three mouse events.
Currently, the "click" action triggers the "click" function directly on the focused (non-select) target element.  This has a couple of key limitations:

1. There is no on-screen indication that a button is clicked or released.
2. Any "click" listener that tries to call `event.preventDefault` will fail.

It would be better to dispatch a "mousedown" and "click" even when the button is pressed, and to dispatch a "mouseup" even when the button is released.

duhrer:
See #91 for details.

### DIFF
--- a/src/js/content_scripts/input-mapper-content-utils.js
+++ b/src/js/content_scripts/input-mapper-content-utils.js
@@ -283,7 +283,7 @@ https://github.com/fluid-lab/gamepad-navigator/blob/master/LICENSE
 
     /**
      *
-     * Click on the currently focused element.
+     * Simulate a mouse click on the currently focused element.
      *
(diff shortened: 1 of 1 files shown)


duhrer:
It seems like directly calling `click` works more consistently than dispatching events.  If I just dispatch `mousedown` and `mouseup` events, sites like [Weavly](https://create.weavly.org/] don't work.  If I also dispatch `click` I get double clicks and odd behaviour with the tab navigation. 

I think I may abandon this one and address the "held button" concerns in the on-screen keyboard using other means.

PR #94, opened 2022-04-07
GH-93: Updated security audit exceptions (resolves #93).

### LINKED ISSUE #93: Updated dependencies to address security failures in CI.
In working on another pull (#92) I noticed that the security audit check in the workflow failed.  We should update dependencies as far as possible anyway, but if there are any leftover or we can't upgrade, we might want to migrate to using npm-audit-resolve for CI.

duhrer:
See #93 for details.


duhrer:
Self-merging so I can use the fix in #93.

Issue #90, opened 2022-04-07
Update chrome extension manifest version.


duhrer:
Loading the plugin now throws an error:

```
Manifest version 2 is deprecated, and support will be removed in 2023. See https://developer.chrome.com/blog/mv2-transition/ for more details.
```

We need to update our manifest version by the end of the year.

dmahajan980:
Hey @duhrer @the-t-in-rtf, I've been away for a very long time and trying to catch up through this issue. Can you please tell if anything was finalised on the bundling part and the major blockers (if any) that are preventing us from settling on a bundler?

duhrer:
Hi, @dmahajan980.  I still need to look at this.  I will also create a ticket to add a privacy policy so that we can get the plugin added to the Chrome Web Store again.

duhrer:
It seems like the new version is now required, so I suspect we'll have to address this before we can restore the web store listing.

duhrer:
After reviewing the approach UIO+ used in migrating and reading [this helpful guide](https://css-tricks.com/how-to-transition-to-manifest-v3-for-chrome-extensions/), I have made a first pass at updating.  Thus far I have confirmed that the following work:

1.  The plugin bundles and loads without errors.
2.  The configuration interface and preferences storage work.

I am in the process of testing all methods.  Thus far I have confirmed the following working (in the order they appear in the configuration menu):

1. click
2. history back button
3. history next button
4. focus on the previous element
5. focus on the next element
6. scroll left
7. scroll right
8. scroll up
9. scroll down
10. switch to the previous tab
11. switch to the next tab
12. close current browser tab
13. close current browser window
14. zoom in on the active web page
15. zoom out on the active web page
16. maximise the current browser window
17. restore the size of the current browser window

The following don't seem to work, I will have to look into that further tomorrow:

1. Opening a new tab
2. Opening a new window.
3. Switching to the previous browser window
4. Switching to the next browser window
5. Re-open the last closed tab or window

duhrer:
I have continued testing and will comment further on the pull.

duhrer:
The work has been reviewed and the pull has been merged.

PR #97, opened 2023-09-26
GH-90: Initial work to migrate to manifest version 3.

duhrer:
See #90 for details.


duhrer:
As mentioned in #90, I still need to figure out a few functions that don't work after the initial conversion.  Submitting the work in progress in case you have time to try it out and comment.

duhrer:
I have everything working except navigating between browser windows.  I will continue working on that.

duhrer:
The window switching now works.  I'll clean up the build failures and update the message listener docs, and then I think this will be ready for review.

duhrer:
This is ready for review.

jobara:
@duhrer did a quick look at the code, seems good; other than the failed CI task. I tried out the extension. Is it expected that it doesn't work on the Google home page?

duhrer:
@jobara, the default built-in starting page in Chrome doesn't get the required scripts injected, which has always been a problem.  We have "start page" logic that's meant to circumvent this by opening the Google URL instead, I'll have to confirm how/whether that's working.

Long term, my idea is to have "new window" and "new tab" open a dialog where you can choose where to go, so that you can't end up on a page you can't navigate away from or control.

duhrer:
The CI failure is a linting error, which I just fixed.

duhrer:
So, previously we use a fluid component for the background script, and it used invoker definitions to change a lot of function signatures.  After more closely reviewing, I discovered that the signature of the wrappers for "new window" and "new tab" was off, resulting in  ([https://www.google.com/](the URL we use by default) not getting passed to the code that creates the new tab or window.

With this fixed, creating a new window (right thumb stick press by default) or creating a new tab (start button press by default) both work as expected, and no longer dump the user in a window/tab they can't control.

duhrer:
Just to confirm, this should be ready for further review.

jobara:
@duhrer I'm still getting some funny behaviour with the browser forward/back where it will kick me back to the google start page in chrome. I'm not exactly sure what happened but I went back and forward fine at first, but then it threw me to the google start page that I can't interact with. What's worse is that even if I type in a URL it will launch that page but after loading it will immediately kick me back to the google start page even without interacting with the gamepad.

duhrer:
Thanks for catching that, @jobara.  I installed the previous release to confirm, and the same thing happens, namely:

1. If you try to go back, you go all the way back, and get bumped back if you manually navigate forward in history.
2. If you try to go forward, you go all the way forward, and get bumped forward if you try to manually go back.
3. If you open a new window and navigate around, you'll stay stuck in whichever direction you were headed in.


I'll write it up as a new ticket and add it to the list of work that needs doing.

duhrer:
Self-merging this and moving on to fixing #98.

Issue #84, opened 2021-08-03
Onscreen Keyboard


Gekocaretaker:
Could there be an onscreen keyboard, for inputs? This way, we don't have to switch to the keyboard whenever we need to type?
Consoles use onscreen keyboards for inputting information.

the-t-in-rtf:
We have been talking about this specifically with regard to #85 (inputting a URL).  In order to use this with things like form inputs, we need to figure out a few things around focus.

If we open our own onscreen keyboard and use focus there to indicate which key will be pressed, does that remove the focus from the form input?  I suspect it would, so we'd have to do something like remember the last focused element and send keyboard events from the onscreen keyboard to that.  This would have the undesirable side-effect of hiding the focused element and interacting poorly when entering text and then navigating away.

One way to avoid this would be to not use focus within the onscreen keyboard, perhaps using a "marker" attribute on the key should be highlighted.

It seems like a good first step to look at pure HTML5/Javascript onscreen keyboards to see if there's anything we can adopt, adapt, or learn from.  @jobara, are you aware of any reasonable browser-based onscreen keyboards?

the-t-in-rtf:
I'm also thinking that we need to decide what keys to reserve for navigation within the onscreen keyboard.  The d-pad keys seem like a good fit for changing position.  For simple use cases, we'd need to reserve a button for clicking a key (perhaps we can reuse the same one that would activate a focused link).

For form inputs, we also need to think about more advanced usage, such as entering capital letters.  One solution would be to have an onscreen "shift" button that only modifies the next key and then turns off, and then a "caps" button that modifies all (letter) keys until it's  turned off.

jobara:
> We have been talking about this specifically with regard to #85 (inputting a URL). In order to use this with things like form inputs, we need to figure out a few things around focus.
> 
> If we open our own onscreen keyboard and use focus there to indicate which key will be pressed, does that remove the focus from the form input? I suspect it would, so we'd have to do something like remember the last focused element and send keyboard events from the onscreen keyboard to that. This would have the undesirable side-effect of hiding the focused element and interacting poorly when entering text and then navigating away.
> 
> One way to avoid this would be to not use focus within the onscreen keyboard, perhaps using a "marker" attribute on the key that should be highlighted.
> 
[...]

PR #102, opened 2023-10-17
GH-100: Initial implementation of arrow key actions (resolves #100).

### LINKED ISSUE #100: Add support for simulated arrow key presses...
In order to implement #84 and #85, we need some means of simulating arrow key presses, so that users can more easily select keys from a grid layout.  This ticket covers adding new (unbound) actions and testing it with the onscreen keyboard.

duhrer:
This pull adds actions to simulate keyboard navigation using arrow keys.  This is a requirement for #84.

See #100 for more context.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,10 @@ The Gamepad Navigator currently supports the following actions for buttons and t
 | **Maximize the current browser window** | Maximize the current browser window. | [maximizeWindow](docs/messageListener.md#messagelistenermaximizewindowleft) |
 | **Restore the size of current browser window** | Restore the size of the currently active browser window. | [restoreWindowSize](docs/messageListener.md#messagelistenerrestorewindowsizeleft) |
 | **Reopen the last closed tab or window** | Reopen the last closed browser tab or window. | [reopenTabOrWindow](docs/messageListener.md#messagelistenerreopentaborwindow) |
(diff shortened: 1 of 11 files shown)


duhrer:
I added functions for analog thumb sticks and buttons.  I've been testing the buttons with the d-pad.

For anyone else who wants to try, here are two pages that support grid-based navigation using arrow keys:

1. [Fluid OSK text input demo](https://duhrer.github.io/demos/fluid-osk/examples/text-input.html)
2. [Launchpad Slide Puzzle](https://duhrer.github.io/demos/react-launchpad-slide-puzzle/)

duhrer:
I've been testing the thumb stick functions with a speed factor of 1.5, which is probably what I'll use when I rework the bindings for #84.

I'm not overly happy with the "cutoff" that we use to prevent "jitter" on noisy sticks.  You still get a few extra scroll actions when the stick bounces back hard, and it's far too easy to trigger the other axis while trying to move along a single axis.

I may introduce a second cutoff for this (eventually these should become preferences, ideally when we rework the configuration panel).

duhrer:
@jobara if you have a moment to try this, I'd appreciate the feedback.  Otherwise I will merge this on my own and continue refining this in the onscreen keyboard work (#84).

duhrer:
I reran the last CI build, it failed with a very suspiciously low-level error (no jQuery) that we've not seen before.

PR #112, opened 2023-11-02
Add "Action Launcher" (resolves #101)

### LINKED ISSUE #101: Add an action to select and execute an arbitrary action from a list of actions.
We have reached the point where nearly all of the buttons/axes on most gamepads are used by the default configuration.  For less commonly used actions, it would be helpful to have a menu to launch an action without having to bind it to a button or axis.  This is particularly appropriate for actions like:

1. Minimising/restoring a window.
2. Opening a new window/tab.
3. Closing a window/tab.
4. Previous/next page in history.
5. Previous/next tab/window.

Although it's tedious, this could also work for the click and tab navigation.  It's not as good a fit for scrolling actions, but these could be handled as though they were responding to a single click from a digital button (like the d-pad).

One thought might be only to present actions which are not bound to an axis/button, but providing the ability to launch any action would be a good start.

duhrer:
See #101 for background.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -2,5 +2,12 @@
     "env": {
         "browser": true
     },
-    "extends": "eslint-config-fluid"
+    "extends": "eslint-config-fluid",
+    "rules": {
+        "indent": [
+            "error",
+            4,
+            { "SwitchCase": 1}
+        ]
+    }
 }
--- a/audit-resolve.json
+++ b/audit-resolve.json
@@ -370,23 +370,51 @@
     },
     "1094304|fluid-lint-all>eslint-plugin-markdown>remark-parse>trim>markdownlint-config-fluid>markdownlint>markdown-it>stylelint>autoprefixer>postcss": {
       "decision": "ignore",
-      "madeAt": 1697531759856
+      "madeAt": 1698935247034
     },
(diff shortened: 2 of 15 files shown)


duhrer:
@jobara, We had one available button left, so button 1 is bound to the new action launcher for testing (this will be cleaned up when I redo the config panel).  I have confirmed that all actions work, including testing the arrow actions with the [onscreen keyboard](https://duhrer.github.io/demos/fluid-osk/examples/).

This is the first use of a new modal that will eventually replace the existing configuration panel (so that the configuration can be updated using the controller itself).  I have tested preserving/restoring focus, tab trapping while open, escape to close, and everything else I could think of.

duhrer:
A little more detail on the relevant default bindings, the shoulder bumpers are basically tab navigation simulators, and button 0 (x on a PS controller) is "click".  You should also be able to entirely operate the modal with those gamepad controls.

duhrer:
OK, I hallucinated adding the button to the default config earlier.  I actually did it now.  If you "Restore Defaults" in the config menu, things will be as I described earlier.

duhrer:
I understand people's time is tight, I will self-merge this and put out a call for wider participation in a QA effort as part of the larger 1.0 release.

Issue #103, opened 2023-10-17
Refactor configuration interface.


duhrer:
Currently, we have a small popup for all supported configuration options, with an idealised picture of a gamepad, and a complex form where you select the input (button or axis) with one drop down, and then select the action to assign to the button.  Once both drop-downs are selected, any additional tool options appear.

This refactor covers:

1. Expanding the popup to a modal dialogue that includes scrollbars.
2. Presenting the current configuration as an editable list of "bindings" (control->action).
3. Presenting overall configuration options like the anti-jitter "cutoff" and the home page that will be used for tabs/windows.

duhrer:
As part of this, we need to politely "defang" the current configuration panel, which appears when the user clicks the gamepad navigator icon.  This should be replaced with a simple page that displays:

1. Information about the name and version of the plugin.
2. A link to the documentation (the repo for now until #114 is done).
3. A button to open the action launcher.
4. A button to open the new configuration panel.

duhrer:
Because there's at least one field (homepage URL) that needs to be edited as text, we need to be able to use this in combination with the onscreen keyboard modal.  If the options panel itself is also a modal, we'd have to manage a transition between two modals.

Since we needed an options menu for #120, I ended up moving this to a standalone page.

Unlike the previous control panel, I included the navigation on the new options panel, which means you can change settings using a gamepad, and can also change to another window or close the window using a gamepad.

PR #122, opened 2023-12-04
GH-84: Work in progress on 1.0 release.

duhrer:
Although I don't have anyone to review the work to date, I'm putting in this pull and self-reviewing before continuing on to #103, which I would like to get help reviewing.


duhrer:
The waters are a little bit muddy here with the initial skeleton for the "new settings" mixed in, but it can't be avoided.  My plan is to post in community forums closer to release and clean up regressions along with anything new discovered in QA.

Issue #125, opened 2023-12-07
Current select handling mangles the UI and doesn't allow editing.


duhrer:
We have special handling for select inputs that opens them by calculating their size and resizing them. This worked previously to allow opening a select, but does not seem to work any longer. I'm not sure it ever really allowed proper navigation either.

Regardless, with the code in progress for the upcoming 1.0 release, it's not possible to navigate through or to choose select box options. This is a blocker for #103 and the 1.0 release, as we cannot add bindings or configure "send key" bindings until this is fixed.

The changes we make to the select element's size are also not undone when we leave the element, which results in mangling the select inputs as we operate them.

To address this, we should add a modal similar to the one we use for text inputs, which allows the user to navigate through the list of options. We should:

1. On clicking a select element, derive a set of choices from the `<option>` tags in the select.
2. Present a navigable list of the choices in a modal with a "Cancel" and "Save" button.
3. Close the modal without selecting a choice when the "Cancel" button is clicked.
4. For single-choice menus, when a choice is selected, move focus to the "Save" button automatically.
5. Enable the "Save" button when one (or more) choices is selected.
6. On "Save", close the modal and change the value of the original select (as we do with text inputs).

Notably, this approach will not help with select boxes in our own modals, as we have no mechanism to transition between modals.  For now we will have to avoid using select boxes in modals, but this seems acceptable.

duhrer:
When building the list of choices, we should display but not allow the user to select `disabled` options.  We should also not display options with the `hidden` attribute.

duhrer:
I'm half wondering if we should try to simulate a click on the selected option, in case there are event handlers on the option.

I also think we should set up a select with a change handler and confirm whether setting the select's value is adequate, or whether we need to trigger the select element's change event ourselves.

duhrer:
I came up with a working solution that can operate our settings panel, and a range of select inputs, including multiple select boxes.

Issue #115, opened 2023-11-10
Refactor default key binding.


duhrer:
Currently, we support a single key binding, which uses the special `_execute_action` command (identical to clicking the plugin icon).

Instead, we should add a simple [command listener](https://developer.chrome.com/docs/extensions/reference/commands/) that will open the action launcher when the key binding is hit.

Issue #109, opened 2023-10-18
Add the option to control the "speed factor" for thumb stick controls.


duhrer:
Currently, we have a per-binding speed factor that controls how frequently a thumb stick setting repeats.  We could add a setting to base the speed factor on the current deflection from neutral, so that pushing a stick all the way to the side repeats faster than a gentle deflection.

We should preserve the existing behaviour as an option as well, as some users may lack the precision to avoid fully depressing the stick, and might prefer a hard-coded and consistent value.

duhrer:
This is slightly more complicated, as there are `frequency` settings (how often) and "factor" settings (how far in one go).  The frequency setting can be incorporated into #108.  The factor setting needs a little review, I'm willing to leave this one open for that work.

Issue #108, opened 2023-10-18
Refactor handling of "discrete" vs. "continuous" controls.


duhrer:
Currently, we have two kinds of controls.  The first are "discrete" controls, that fire when the control value is above the cutoff.  Because there is no checking for "already on", these controls can be called repeatedly (and rapidly) when an analog button like a trigger is used.

The second are "continuous" controls, which poll for updates and keep firing the action as long as the value is still above the cutoff threshold.  These can only be bound to "axes" at the moment.  In cases like scrolling, the value is also used to determine how hard/fast to move, so that gentle thumb stick deflection scrolls slowly, and full deflection scrolls quickly.

For both types of action, each action models its own value handling, which results in a lot of duplicated code and inconsistent behaviour.  We should  refactor such that:

1. Each type of control extends a parent grade that enforces the basic contract (jitter cutoff in the case of "discrete" controls, for example, and polling in the case of "continuous" controls).
2. Both types of actions are available for all controls, such that:
    - Buttons can be used with "continuous" controls. For example, we should be able to scroll repeatedly by holding a d-pad directional button, or navigate continuously by holding a bumper.
    - Thumb sticks and triggers can be used with "discrete" actions without constantly repeating.

This would be a good time to standardise the contract for actions in general, so that we don't have a method signature with every possible parameter, and instead pass an action options object for each call.

duhrer:
The supported parameters should be part of the new action contract.  Previously we've used things like [JSON schemas](https://json-schema.org/) to model parameters, which allows both input validation and UI generation.

Given that the range of options is somewhat more constrained here, we might try for a simpler definition, which describes:

1. Which fields an action supports.
2. The allowable values for the field (i.e. boolean vs. number, maximum and minimum values).
3. The default value for the field.

We could also use our [FSS variation on JSON schemas](https://github.com/fluid-project/fluid-json-schema/issues) informally, i.e. without using the validation infrastructure, i.e. only using it for UI generation.  We could add a schema-linting step to our tests as a safety check, without having to inject AJV and a bunch of fluid-json-schema code into the page.

duhrer:
The following should only support "discrete" operation:

1. `openActionLauncher`
2. `openSearchKeyboard`
3. `openNewWindow`
4. `openNewTab`
5. `closeCurrentTab`
6. `closeCurrentWindow`
7. `reopenTabOrWindow`
8. `maximizeWindow`
9. `restoreWindowSize`
10. `click`

[...]

PR #130, opened 2023-12-13
Major refactor for 1.0 release.

duhrer:
This PR includes (hopefully the last) major refactoring work in preparation for the 1.0 release. In summary:

1. Completely new configuration interface, which is controllable using the gamepad.
2. New default bindings.
3. There is now an onscreen keyboard to allow text input.
4. There is now a search action.
5. There is now an "action launcher" to access most actions with a single button.
6. Select elements are now actually usable using the gamepad.
7. Most controls can now repeat, this is configurable.
8. The gamepad polling and repeat rates are user configurable.
9. The "new page" URL is configurable.
10. Vibration is now supported when an action cannot be completed (this is configurable).
11. There are many more guardrails to prevent navigating to a window/tab that can't be controlled.
12. Updated the documentation.
13. Added additional tests and fixed previously undetected bugs.

Includes work on: #59 #103 #106 #108 #109 #115 #125


duhrer:
Last failure was because of linting errors (new dev environment without the pre-commit linting hooks, fixed now).

Issue #162, opened 2024-02-16
Focus styles not applied consistently.


duhrer:
We now have a clearer explanation for why `:focus-visible` styles aren't always applied when we use bumper navigation (see #161).

However, the problems described in #161 don't explain why I can occasionally get into a state where modals don't display focus.  All of those styles use `:focus` (in fact we don't use `:focus-visible` anywhere in our code base).

This ticket tracks research into modal focus problems like this.

duhrer:
The next time I get into a broken state, I'd like to:

1. Set a break point after we programatically focus to see what `document.targetElement` is.
2. Examine the state of the element that should have been focused.
3. Try tabbing or otherwise using the keyboard to see if it unsticks things (this would suggest that this is a variation on #161).

duhrer:
In further testing for #161, I discovered that:

1. Some of the elements for which focus is not displayed have `:focus` rather than only `:focus-visible` styles.
2. `document.activeElement` is being set correctly when triggering focus.

This suggests to me that this is indeed the same issue.  Leaving this open until I can get the menus in a state to confirm.

duhrer:
I think I'm going to close this one once #164 is merged.

Issue #161, opened 2024-02-15
Visible focus is not shown for all elements.


duhrer:
I've been occasionally seeing issues where focus indicators would be visible for some but not all elements.  It was clear that focus was still changing, you could see the URL of a focused link in the footer of the browser window, but the focused element didn't have any styles. Hitting keyboard generally "unsticks" things and focus is visible on all elements.

I now have a consistent way to reproduce the stuck state in at least one circumstance:

1. Open a new window (with the Google search).
2. Enter text in the search field.
3. Navigate to the search button and click it.
4. Use bumper navigation to try and reach the search results.

The buttons for the type of search ("images", "shopping", etc.) have focus styles, but when you reach the results, no focus styles are applied, even though it's clear that focus is changing, and you can click the focused search result.

duhrer:
I've seen variations on this where the modals would lose focus until you reload or hit a tab on the keyboard as well, but I can't consistently reproduce that.

duhrer:
Although sadly we can't operate the code with the Gamepad Navigator because it's in an iframe (see #155), [this CodePen](https://codepen.io/cagerattler/pen/poQqPBq?editors=1111) illustrates what I think is the core of the problem, namely that `:focus-visible` styles stop being applied if there's not actual keyboard input.

duhrer:
There are a few threads like this, but [here's an example Stack Overflow thread](https://stackoverflow.com/questions/69283396/what-can-cause-focus-visible-style-to-not-be-applied-when-using-element-focus) that suggests working around the problem by making the element editable before focusing, and then changing it back after.

duhrer:
Looks like there is [a configuration option to use with `focus()`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/focus) to indicate that visible focus should be triggered.

duhrer:
I can't get any of the hacks or the optional parameter to work.

Changing the browser preferences does work, i.e.  visit `chrome://settings/accessibility`, then toggle "Show a quick highlight on the focused object".

Depending, maybe we can detect if that's not set and advise people to change it.

duhrer:
Looks like an extension can change this using the [`accessibilityFeatures`](https://developer.chrome.com/docs/extensions/reference/api/accessibilityFeatures) parts of the web extensions API.

One option would be to add actions to enable/disable that setting, or perhaps one to toggle the setting.

duhrer:
[...]

PR #164, opened 2024-02-21
GH-162: Initial implementation of overlay-based 'focus fix'.

duhrer:
I had been intermittently seeing cases where focus styles were not displayed, this pull attempts to provide a workaround for the underlying issues.

See #162 and #161 for more details on the research and the browser behaviour that required the workaround.

### DIFF
--- a/audit-resolve.json
+++ b/audit-resolve.json
@@ -459,6 +459,10 @@
     "1096305|grunt-prompt>inquirer>lodash": {
       "decision": "ignore",
       "madeAt": 1706777084332
+    },
+    "1096487|grunt-prompt>inquirer>lodash": {
+      "decision": "ignore",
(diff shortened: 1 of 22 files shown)


duhrer:
As mentioned in the issue, the easiest way to see the underlying problem is to launch a search with the action launcher.  Bumper navigating through the results does not display anything for multiple "bumps", and then finally only displays the destination URL for focused links.

With the new setting on, it's much clearer that menu elements, etc. are focused during this "invisible" period.  There are two ways to enable the new setting:

1. Open the settings panel, it's the last entry under "General Preferences".
2. This is the first and only setting with a dedicated action to "toggle" it, so you can also open the action launcher and click "Toggle optional focus indicator".

Issue #166, opened 2024-02-23
Improve styles for settings panel


duhrer:
In working on #162, I added a new preference, and noticed that the styles for each preference are way too specific.  We need a  consistent style for the preference label, description, and body, and not individual classes for every preference.  The new "focus fix" preference uses the general styles and class name that all of them should.

Once this is done, we can improve the styles and markup.  Currently, the label and body are displayed as columns, which means the left edge of the body is not consistent for each preference.

Instead, the label should be displayed above the body, and the body should be slightly indented.  This will require a bit more vertical room, but should be a lot more consistent.  It should also only require changing the styles themselves.

PR #167, opened 2024-02-24
GH-166: Cleaned up styles and UI grades for settings panels.

duhrer:
See #166 for details.

In summary, the UI grades in the settings panel were not cleanly divided, the low-level UI elements like toggles and such were also managing the labels and descriptions.  I pared down the simple grades to just handle their own UI and made cleanly expressed wrapper grades for the sites that use them.

I also noticed that a lot of very similar elements were using overly specific selectors, there should have been one class for binding labels, but there were four or five.  I cleaned this up and also improved the naming in general.

At the end I went through with the coverage tool in the Chrome dev tools and pared down unused styles following the refactor.

As a side effect, while I was working on this I found a bug lingering from some of the focus refactoring, where removing a binding would throw an error when attempting to focus.  I fixed this.


duhrer:
This also fixes the misalignment of the focus styles with the binding options icons.

